PR #4, opened 2022-04-10
update docs for status message

think-in-universe:
We need to initialize the contract with `init()` method before call `set_status(message)`, otherwise the `Contract state is empty` error will be thrown by jsvm.


ailisp:
Thanks! Yes I forgot to document this step. How do you figure out :)

think-in-universe:
> Thanks! Yes I forgot to document this step. How do you figure out :)

I had a try the **status-message** example yesterday and encountered the error. I took a look at the generated `status-message.js` and noticed the generated `init()` method was never used 😃

PR #13, opened 2022-05-10
Integrate with workspaces-js 2.0 example

ailisp:
Example written and passed for status-message. Can be use as a reference test setup for other examples and user projects


volovyk-s (approved):
Looks cool! We will need to add some helpers to Workspaces JS or to the SDK with time. But it's good for now.

ailisp:
> Looks cool! We will need to add some helpers to Workspaces JS or to the SDK with time. But it's good for now.

Yes. Do you think `encodeCall` should better be placed in workspaces-js? I don't find a proper place for it in near-sdk-js, since `sdk` package is meant to imported in contract (quickjs).

ailisp:
I've tested the tests still work after #12

volovyk-s:
> > Looks cool! We will need to add some helpers to Workspaces JS or to the SDK with time. But it's good for now.
> 
> Yes. Do you think `encodeCall` should better be placed in workspaces-js? I don't find a proper place for it in near-sdk-js, since `sdk` package is meant to imported in contract (quickjs).

Maybe we will create a separate package, not sure. Let's design that after releasing SDK JS on NPM.

PR #14, opened 2022-05-10
RAM usage limited

volovyk-s:
- `make -j` is killing PC if it has 16 or fewer Gb of RAM.
I think that the standard setup should limit the number of jobs and work on most machines.
People with powerful computers can change it during development if they want.

- added `yarn.lock` files to make the setup and builds more stable

`make` doc:
```
-j [N], --jobs[=N]          Allow N jobs at once; infinite jobs with no arg.
```


setup.sh
```diff
@@ -19,7 +19,7 @@ tar xvf wasi-sdk-11.0-"${system}".tar.gz
 # binaryen
 git clone --branch wasi-stub --single-branch https://github.com/near/binaryen
 cd binaryen
-cmake . && make -j
+cmake . && make -j 8
```
ailisp, line 22: Is there developers with less than 4 core 8 threads? If we need to consider those developers, maybe simply drop `-j`. `make` is slower than `make -j 8` or `make -j`, but it'll work on slow/less RAM computers. `setup.sh` is a one time thing, slower is okay.

```suggestion
cmake . && make
```

ailisp (approved):
Looks good! Will also need to update status-message/yarn.lock after #13

PR #16, opened 2022-05-10
Instraction to install deps with yarn added

volovyk-s:
Small, but important detail.


ailisp (approved):
It is a working solution for now! After making sdk folder a js package, we may add it as a dependency instead of `import ../../sdk` from example (my bad hack)

PR #17, opened 2022-05-11
cross-contract-call example

volovyk-s:


ailisp (approved):
Nice job! It's our most complicated example so far!

PR #19, opened 2022-05-13
Fix the bug that after cross contract call, jsvm_storage_write write to callee contract

ailisp:
This bug result in state not updating after a cross contract call (actually it's even worse, it wrote to the wrong contract). After the fix, state write after cross contract call is writing to the correct place (caller contract). 

Also fix several api function, where it returns BigUint64 (JS BigInt) instead of Number.

And add stringify and parse to cross contract call args to simplify developer's work. Developer can now do:
```
let available = near.jsvmCall('status-message.test.near', 'get_status', [accountId])
// available is "AVIALABLE" instead of '"AVAILABLE"'
```

~Also optimize to reuse the quickjs instance to do cross contract call. Should save some wasm instructions and therefore gas.~
Update: we can't do this, because otherwise user can modify `env.` host functions, then reuse quickjs would use modified version, cause security attacks.

### DIFF
--- a/jsvm.c
+++ b/jsvm.c
@@ -944,16 +944,15 @@ static JSValue near_storage_has_key(JSContext *ctx, JSValueConst this_val, int a
 
 static void js_add_near_host_functions(JSContext* ctx);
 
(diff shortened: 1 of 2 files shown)


volovyk-s (approved):
Cool! Now it works :)

PR #24, opened 2022-05-18
Non-fungible token example

itegulov:
Yet to write tests, but tested manually on testnet

Also had to use LookupMap for now, but ideally we should use TreeMap


ailisp:
> Yet to write tests, but tested manually on testnet
Also had to use LookupMap for now, but ideally we should use TreeMap

Sounds good! I'm still writing TreeMap, which depends on Vector. Vector is mostly ready.

ailisp (approved):
The example is good! I updated some import/path to use near-sdk-js npm package made by Serhii

PR #25, opened 2022-05-18
Add on-trie vector data structure

ailisp:
Testing is more in a simulated way in quickjs repl. Will need to think of a better way to test and also use it in status-collections example


src/collections/vector.js
```diff
@@ -0,0 +1,128 @@
...
+        return this.length == 0
+    }
+
+    indexToKey(index) {
+        // TODO: there's no efficient way to convert a Uint64 (BigInt) to Uint8Array
+        // we limit index to be Uint32 for now
+        let data = new Uint32Array([index])
+        let array = new Uint8Array(data.buffer)
+        let key = u8ArrayToString(array)
+        return prefix + key
```
austinabell, line 29: shouldn't this be 
```suggestion
        return this.prefix + key
```
or am I misunderstanding this?

src/collections/vector.js
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    hasNext() {
+        return this.current < this.vector.len()
+    }
+
+    next() {
+        let value = this.vector.get(this.current)
+        this.current += 1
+        return value
```
austinabell, line 126: Isn't the JS iterator standard return an object with `value` and `done` fields? Does that not fit within the SDK?

src/collections/vector.js
```diff
@@ -0,0 +1,128 @@
...
+        return this.length
+    }
+
+    isEmpty() {
+        return this.length == 0
+    }
+
+    indexToKey(index) {
+        // TODO: there's no efficient way to convert a Uint64 (BigInt) to Uint8Array
+        // we limit index to be Uint32 for now
```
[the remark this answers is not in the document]
ailisp: I was referencing to old collections/Vector instead of new store/vec, which is using u64. Because examples are using `collections/*` and we're trying to achieve features parallel to near-sdk-rs 1.0 or 2.0. Since new one has moved to u32, it's good now!

src/collections/vector.js
ailisp: Good catch! This is how repl manual test can get wrong 😂

src/collections/vector.js
ailisp: TIL that! I think it should fit, let me try

ailisp: Got it work

PR #38, opened 2022-05-18
feat: Added Makefile for simpler build process

ChaoticTempest:
Added the `Makefile` for a simpler build process for users trying to build the SDK.

@ailisp I never got the reason why we went with C, since I was thinking we could've reused the Rust `near-sys` crate to bind the host functions instead of rewriting it all in C. There's probably a reason behind it, but if we continue to go down the route of using C, we should have a proper build system such as cmake or meson, so people can build the SDK more readily instead of having to manually install dependencies and run into reproducibility issues


ailisp:
@ChaoticTempest jsvm.wasm is essentially a JavaScript runtime plus injected near host functions. The JavaScript runtime is modified QuickJS, which is written in C. To inject near host functions to QuickJS, it also need to be done in C

ailisp (approved):

ailisp:
Looks good to me, I update it with @volovyk-s 's new project layout

volovyk-s (approved):
Nice @ChaoticTempest!  Thanks for the update @ailisp .
It's already good for merging, but we will need to add the ARM (M1) build artifact to the res folder (I will need your help with that).

volovyk-s:
@itegulov thanks for the help!

PR #49, opened 2022-05-19
Readme Updates

BenKurrek:
outlined differences between wasm and enclave versions and provided instructions on how someone can start using the enclave version

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ There are pros and cons of each method but allow the user to write smart contrac
 
 ## WASM Version
 
-To start using the WebAssembly version, checkout the `standalone` branch and follow the instructions in the README.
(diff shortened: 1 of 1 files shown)


ailisp:
While the documentation you made is good, I think the product decision was not to market and maintain standalone branch of near-sdk-js. Sorry haven't made this clear before!

/cc @AustinBaggio

gagdiez:
@ailisp we have a couple of discussions ongoing on why it is beneficial to have both versions. In a nutshell, it is nice if users can choose their own pathway:

1. Release the standalone wasm file paying for storage, and use it as just-another-contract
2. Release the non-standalone wasm file in the js-vm and be: isolated from other contracts, pay for every method call, deploy/update contracts in a different way from RUST/AS, interact from near-cli and near-api-js in a different way from RUST/AS

But I do think that pulling another branch and following a different README is a bit too convoluted. We could modify `near-sdk build` to add a flag for standalone wasm files.

PR #52, opened 2022-05-20
add lookup set

ailisp:


volovyk-s (approved):
I already feel that we need unit tests :) let's prioritize it some day

PR #54, opened 2022-05-20
chore: adjust raw `near call` to newer `near js` calls in README

ChaoticTempest:
Converted over most of the `near call ...` into `near js ...` ones. I also corrected some IDs that didn't look right, so let me know if those make sense.

Also, hid the output in html tags so that the README isn't too long and cognitive load becomes a problem when a new user comes to read it

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,27 @@ near-sdk-js (master) near js deploy --accountId test.near --base64 hello_near.ba
     To see the transaction in the transaction explorer, please open this url in your browser
     http://localhost:9001/transactions/Df7txPSFWwaBLTz61pSxoVrPPu6qY7fUTJ31xuQtXDBf
     ''
+
+</p>
+</details>
+
+```sh
+near-sdk-js (master) near js call test.near hello --accountId test.near --deposit 0.1 --jsvm test.near
+```
+
+<details>
+<summary><strong>Or use the raw CLI call instead</strong></summary>
+<p>
+
     near-sdk-js (master) near call test.near call_js_contract --accountId test.near --base64 --args $(node encode_call.js test.near hello '')
+
+</p>
+</details>
+
+<details>
+<summary><strong>Example output</strong></summary>
+<p>
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -46,109 +75,215 @@ target/debug/neard run
...
-Starting deployment. Account id: test.near, node: http://localhost:3030, helper: http://localhost:3000, file: jsvm.wasm
-Transaction Id EGVd29tthMp7fqkDgP8frftgZhhb3FVazaFhvXpYXNhw
-To see the transaction in the transaction explorer, please open this url in your browser
-http://localhost:9001/transactions/EGVd29tthMp7fqkDgP8frftgZhhb3FVazaFhvXpYXNhw
-Done deploying to test.near
+near-sdk-js (master) near deploy test.near jsvm.wasm
 ```
 
+<details>
+<summary><strong>Example output</strong></summary>
```
ailisp, line 84: Nice I like the example output! Much cleaner than before

examples/cross-contract-call/README.md
```diff
@@ -21,26 +21,54 @@ yarn test
...
 ## Call the contract
 Under the root dir of near-sdk-js, call the `set_status` and `set_person_on_call` using this pattern:
 
+```sh
+near js call <contract-id> <function-name> [--args '<parameter>'] --deposit 0.1 --accountId <signer-id> --jsvm jsvm.test.near
 ```
-near call jsvm.test.near call_js_contract --accountId <accountID> --base64 --args $(node encode_call.js <contract-account-id> <function-name> '[<parameter>]') --deposit 0.1
-```
+
+or with t<details>
```
ailisp, line 67: typo here?

ailisp: Also suggest to drop the raw commands in these example READMEs. I think just let advanced user know about the raw command in one place?

ailisp (approved):
Thanks, it's a comprehensive update! Except the typo, the other comments are just my opinions, feel free to merge it if you think it's ready.

volovyk-s (approved):
Awesome! We will need to reorg README in the future. Move low-level APIs to a separate file, add simple contract examples, reorg examples READMEs, etc. But no blockers here!

PR #56, opened 2022-05-23
Add unordered map

ailisp:


ailisp:
Merged for now, will have this test in status-message-collections example

Issue #51, opened 2022-05-20
Figure out a way to unit test collections


ailisp:
quickjs doesn't support require, so cannot conditionally require mocked host functions to test collections

We don’t have to have this before the first release, but certain collections' functionality need at least being tested indirectly by examples

ailisp:
The other attempts I made or thought of:

- Find how to let the js know about whether it is running on nodejs or quickjs. Haven't found a way
- Tried to let the js file directly run in quickjs for test. We'll need to write raw tests since most test framework would not in quickjs environment
- Tried to use require to import different file at runtime. However quickjs doesn't support it.
- Tried to use import as nodejs import syntax, error.
- Tried to use import as function in quickjs, doesn't give error, but it always return empty object
- Tried to mutate a module after import, but it is frozen

Some ways that might work:
- Config babel in a smart way, so that it compiled using mocked host functions for test and resulting module can be tested with quickjs binary
- Modify quickjs.c, add runtime conditional import support
- Modify jsvm.c, make it conditional compiled. For test version, all host functions are mocked

volovyk-s:
If all the above options will fail - we can use workspaces for that.

ailisp:
Yes, workspace is a good approach. We can write contract to test specific APIs

PR #58, opened 2022-05-24
Add status message collections example

ailisp:
This also tested basic functionality of LookupSet and UnorderedMap (therefore also indirectly Vector), but not full. If intend to test all the features, it doesn’t quite fit into the status message example. So we would either have separate workspace test to test high level APIs, or test it in a better way as discussed in #51


volovyk-s (approved):
Cool!

PR #62, opened 2022-05-26
Unit tests setup added (workspaces approach)

volovyk-s:
Known limitations:
- We need to wrap wrappers for all the APIs
- For now we are forced to have everything in one contract or add more test contracts. We need to improve `near-sdk build command`. It should accept the contract path and be able to build several contracts.


ailisp:
Thanks! This helps me a lot and I should be able to add all other tests

ailisp (approved):

PR #66, opened 2022-05-31
Collection tests

ailisp:


volovyk-s (approved):
Awesome!
Maybe we will need to add something like `near-sdk build src/` in the future.

PR #75, opened 2022-06-01
Borsh

### LINKED ISSUE #242: borsh support in near-sdk-js

volovyks:
It is throwing errors for now. Trying to find the reason.

The difference between JSON and borsh serialization is 1 line of code + schema. It will be great to have this logic in NearContract in the future.


gagdiez:
Interesting, it behaves as if it was not finding the exports of borsh... could it be because of how the borsh.js was compiled/created?

ailisp:
I made an attempt to  fix rollup configuration to use borsh, now import serialize/deserialize from borsh works. Now it fails for a different issue: borsh uses `Buffer`, but that's a nodejs API that quickjs doesn't have

gagdiez:
Could we implement a Buffer object as a wrapper of uint8array and import it globally?

ailisp:
@gagdiez Cool! A good step towards success! Let me further explore on your approach

ailisp:
I'm able to make it compile too, by using https://github.com/feross/buffer, update cli.js to config rollup, use more gas since compiled contracts become bigger.

And mutate the `status-message/node_modules/borsh/lib/index.js`, add this import:
```
var Buffer = __importDefault(require('buffer/')).Buffer
```

The borsh test now fail of this error:
```
Smart contract panicked: cannot read property \'alloc\' of undefined\\n    at BinaryWriter (build/status-message-borsh.js:6816)\\n    at serialize (build/status-message-borsh.js:7118)\\n    at serialize (build/status-message-borsh.js:7246)\\n    at NewClass (build/status-message-borsh.js:42)\\n    at init (build/status-message-borsh.js:7263)\\n
```

Looking at build/status-message-borsh.js, it seems correctly included https://github.com/feross/buffer, but it still doesn't work, because feross/buffer doing this check:

```
	    if (typeof window !== 'undefined' && typeof window.Buffer !== 'undefined') {
	      Buffer = window.Buffer;
	    } else {
	      Buffer = buffer.Buffer;
	    }
```

And in quickjs environment `window` is undefined. Even if we fixed this check,  borsh-js would still fail, because feross/buffer doesn't have `Buffer.prototype.alloc` function. Therefore, what we need is a **quickjs** buffer implementation, we can start from feross/buffer, only standard ES2020 features can be used and implement at least what borsh used.

ailisp:
test-template is working locally, strange that it doesn't work on CI 🤔

volovyk-s:
- It is failing on my machine as well.
- Borsh example tests work (we need to add it to CI after merging master).
- Setup should be the same in all the examples. We can achieve this by merging the master to this branch since it has only one test project now. I have tried to do it, but it led to several other problems.

examples/template/__tests__/test-template.ava.js
```diff
@@ -24,8 +24,8 @@ test.beforeEach(async t => {
     // Deploy and init the counter JS contract
     const counter = await root.createSubAccount('counter');
     let contract_base64 = (await readFile('build/contract.base64')).toString();
-    await counter.call(jsvm, 'deploy_js_contract', Buffer.from(contract_base64, 'base64'), {attachedDeposit: '400000000000000000000000'});
-    await counter.call(jsvm, 'call_js_contract', encodeCall(counter.accountId, 'init', []), {attachedDeposit: '400000000000000000000000'});
+    await counter.call(jsvm, 'deploy_js_contract', Buffer.from(contract_base64, 'base64'), {attachedDeposit: '800000000000000000000000'});
```
volovyk-s, line 27: The contract is 37! times bigger now :)
I think it's better not to support Borsh at all. But I believe we will be able to find the solution.

examples/status-message/package.json
```diff
@@ -5,15 +5,18 @@
...
     "build:simple": "near-sdk build src/status-message.js build/status-message.base64",
     "build:collections": "near-sdk build src/status-message-collections.js build/status-message-collections.base64",
+    "build:borsh": "near-sdk build src/status-message-borsh.js build/status-message-borsh.base64",
     "test": "ava"
   },
   "author": "Near Inc <hello@nearprotocol.com>",
   "license": "Apache-2.0",
   "dependencies": {
-    "near-sdk-js": "file:../../"
+    "borsh": "https://github.com/near/borsh-js/tarball/import-buffer",
```
volovyk-s, line 17: I do not understand how this branch is different and why are we using it here. Can we use `fix-build` branch instead? (or maybe master if we will merge it).

package.json
```diff
@@ -29,11 +29,11 @@
   "dependencies": {
     "@babel/core": "^7.17.5",
     "@babel/plugin-proposal-decorators": "^7.17.2",
-    "rollup": "^2.61.1",
-    "rollup-plugin-sourcemaps": "^0.6.3",
     "@rollup/plugin-babel": "^5.3.1",
-    "@rollup/plugin-commonjs": "^21.0.1",
+    "@rollup/plugin-commonjs": "^22.0.0",
```
volovyk-s, line 33: This change helps to fix `status-message-collections`, but I have no idea why :/

cli/cli.js
```diff
@@ -58,10 +59,10 @@ async function build(argv) {
     const bundle = await rollup({
         input: SOURCE_FILE_WITH_PATH,
         plugins: [
-            nodeResolve(),
+            nodeResolve({preferBuiltins: false}),
             sourcemaps(),
-            // commonjs(),
-            babel({ babelHelpers: 'bundled' })
+            babel({ babelHelpers: 'bundled' }),
+            commonjs()
```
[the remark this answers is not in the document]
ailisp: commonjs is used for import non-es packages, if I didn't remember it wrong, import commonjs module doesn't work without it.

examples/status-message/src/status-message-borsh.js
```diff
@@ -0,0 +1,74 @@
+import { NearContract, NearBindgen, call, view, near } from 'near-sdk-js'
+import { serialize, deserialize } from 'borsh';
+import { panic } from '../../../src/api';
+import { Buffer } from 'buffer/'
```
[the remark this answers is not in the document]
ailisp: Yes, should do so

examples/template/__tests__/test-template.ava.js
ailisp: Yes this is how commonjs module worse than esm, rollup cannot handle just importing the used functions. And you found it's even worse that it also packaging all borsh code :(

examples/status-message/package.json
ailisp: It has a `import {Buffer} from `buffer/` in borsh.ts, and borsh.js built from that. Without this line it doesn't work.

package.json
ailisp: Maybe some bug fixes happened in 22.0?

ailisp:
> It is failing on my machine as well.

Perhaps due to commonjs increase template.js size (it depends on lodash). But I still didn't figure out why it passes on my machine, even with a clean build.

> Borsh example tests work (we need to add it to CI after merging master).

Good to hear. Yes.

> Setup should be the same in all the examples. We can achieve this by merging the master to this branch since it has only one test project now. I have tried to do it, but it led to several other problems.

Got it. Is the problem only related to template and status-message-borsh?

ailisp:
Status summary: we made borsh work by a series of hacks in this PR. But enabling commonjs modules enlarge the contract size significantly. We plan to ship borsh also in esm format to resolve

ailisp:
Closed as completed by #361

PR #76, opened 2022-06-01
Status examples merged

volovyk-s:
We have a lot of redundant boilerplate code in examples. Especially, if we will add one more example that implements `borsh` serialization.


ailisp (approved):
Good merge!

PR #77, opened 2022-06-02
Allow to pass arguments as a dictionarry

volovyk-s:
Before:
```JS
await ali.call(jsvm, 'call_js_contract', encodeCall(testContract.accountId, 'set_values', ['newVal1', 'newVal2', 'newVal3']), { attachedDeposit: '100000000000000000000000' });
```
After:
```JS
await ali.call(jsvm, 'call_js_contract', encodeCall(testContract.accountId, 'set_values', { param1: 'newVal1', param2: 'newVal2', param3: 'newVal3' }), { attachedDeposit: '100000000000000000000000' });
```

### DIFF
--- a/tests/__tests__/function-params.ava.js
+++ b/tests/__tests__/function-params.ava.js
@@ -1,0 +1,65 @@
+import { Worker } from 'near-workspaces';
+import { readFile } from 'fs/promises'
+import test from 'ava';
+
+function encodeCall(contract, method, args) {
+    return Buffer.concat([Buffer.from(contract), Buffer.from([0]), Buffer.from(method), Buffer.from([0]), Buffer.from(JSON.stringify(args))])
+}
+
+test.beforeEach(async t => {
+    // Init the worker and start a Sandbox server
+    const worker = await Worker.init();
+
+    // Prepare sandbox for tests, create accounts, deploy contracts, etx.
+    const root = worker.rootAccount;
+
+    // Deploy the jsvm contract.
+    const jsvm = await root.createAndDeploy(
+        root.getSubAccount('jsvm').accountId,
+        '../res/jsvm.wasm',
+    );
+
+    // Deploy test JS contract
+    const testContract = await root.createSubAccount('test-contract');
+    let contract_base64 = (await readFile('build/function-params.base64')).toString();
+    await testContract.call(jsvm, 'deploy_js_contract', Buffer.from(contract_base64, 'base64'), { attachedDeposit: '400000000000000000000000' });
+    await testContract.call(jsvm, 'call_js_contract', encodeCall(testContract.accountId, 'init', []), { attachedDeposit: '400000000000000000000000' });
+
+    // Test users
+    const ali = await root.createSubAccount('ali');
+    const bob = await root.createSubAccount('bob');
+    const carl = await root.createSubAccount('carl');
+
+    // Save state for test runs
+    t.context.worker = worker;
+    t.context.accounts = { root, jsvm, testContract, ali, bob, carl };
+});
+
+test.afterEach(async t => {
+    await t.context.worker.tearDown().catch(error => {
+        console.log('Failed to tear down the worker:', error);
+    });
+});
+
+
+// test('Developer can pass parameters as dictionary', async t => {
+//     const { ali, jsvm, testContract } = t.context.accounts;
+
+//     await ali.call(jsvm, 'call_js_contract', encodeCall(testContract.accountId, 'set_values', ['newVal1', 'newVal2', 'newVal3']), { attachedDeposit: '100000000000000000000000' });
+
+//     t.is(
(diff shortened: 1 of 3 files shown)


BenKurrek:
This is great - definitely a good change. I'll take a look at the PR tomorrow :)

examples/clean-state/src/index.js
```diff
@@ -3,17 +3,17 @@ import { NearContract, NearBindgen, call, view, near, LookupMap } from 'near-sdk
 @NearBindgen
 class CleanState extends NearContract {
     @call
-    clean(keys) {
+    clean({keys}) {
```
ChaoticTempest, line 6: Is it possible to modify each method, and replace them with one that already inlines `{` `}` between all the arguments? We can get rid of forcing the users to constantly write these brackets otherwise and would make the classes look a lot simpler

ailisp: It is possible as we have full access to AST. However I think it's controversial to do so - it's a hidden magic turns the user contract to non standard JavaScript.

ailisp:
I think this is a great and smart approach! it is true that's more flexible and extendable. Also you reused JS's language feature, should be natural and no surprise to the developer. This approach also reduce the complexity of my proposed approach: walk the AST to record param names and generate code to map dict-arguments into positional argument. So I think your approach is better!

examples/clean-state/src/index.js
ChaoticTempest: hmm, but would there be a case where the contract is used as standard javascript though? We don't have any unit tests, and workspaces just makes RPC calls anyways. And the thing is, it's all very magicful anyways when we're sending the arguments across the wire onto the network. Plus if we ever decide to support borsh, the brackets `{}` would make it seem weird in that case

volovyk-s: During contract development you are getting errors. The first place you go is *contract".js file, because error givese you line where it happened. Long magical code will be unfortunate here.

volovyk-s: Not sure how it will affect borsh.

gagdiez:
loving the idea. I think we **MUST** force users to use named parameters, in that way, if I forgot to pass a parameter, the contract will panic with a meaningful error. Passing values by parameter is highly error-prone, specially in JS, where values are not typed.

examples/clean-state/src/index.js
ChaoticTempest: >  The first place you go is *contract".js file, because error givese you line where it happened. Long magical code will be unfortunate here

aren't we generating a separate contract.js file anyways? So error output (or the line the error occurred) would just match up with the generated file then or am I missing something? The generated file should have these inplaced `{}` too so I don't see an issue there.

> Not sure how it will affect borsh

Not entirely sure how the structure of borsh could like, but the brackets `{}` would match more closely to JSON, while borsh could just be any plain old object that's not necessarily a dictionary. Depends on how we want to serialize/deserialize such an object

ailisp: > Plus if we ever decide to support borsh, the brackets {} would make it seem weird in that case

When you make this kind of signature:
```
function rand({ param1, param2 })
```
argument encoding with borsh, it can't work since the order of param1 and param2 is not guaranteed. You can use a borsh-encoded hashmap which has param names and values, but that violates to you want a dense, succinct benefit from borsh. So practically it has to be:
```
function rand(param1, param2)
```

Therefore we have named parameter for json, and non named parameters for borsh. With Serhii's design, both can be achieved by expressing in standard JS:
```
function rand({ param1, param2 })
function rand(param1, param2)
```

If we modify code generation to auto add `{` `}`, user cannot achieve the later case.

ailisp: Good points.

```
@call
@borsh
function rand(param1, param2) {}
```

seems quite good.

> If we do decide to support such a thing in the future, I think the syntax of {param1, param2, ...} might be unwieldy to specify which parameters have what serialization format, but we can think about that later since we won't stabilize 1.0 quite yet

The problem is that, even with above `@borsh` you cannot specify per paramter serialization - js decorators can only on function/method/class, not parameters.

Also I want to know how useful is per-param serialization, e.g.:
```
fn rand(&self, #[serializer(json)] param1, #[serializer(borsh)] param2) {}
```
What is the use case of it, is that only happen at callback function of a borsh contract cross contract call a json contract?

ailisp (approved):
Looks great, I think this is the best solution we can achieve before the consensus. Whether this or @ChaoticTempest 's suggest is better is controversial, let's follow up the discussion after this PR

examples/clean-state/src/index.js
ailisp: Assume we want per-param serialization, given JS decorators' limitation:  can only on function/method/class, not parameters. we possibly do this:
```
@call
@customSerializer({'param2': 'borsh'})
function rand(param1, param2) {}
```
In this case either `rand(param1, param2)` or ``rand({param1, param2})` should work. So Serhii's approach is still good

Issue #79, opened 2022-06-03
TypeScript support tracking issue


austinabell:
Scoping this out, seems totally possible for this lib to be converted to typescript.

Unsure of the overhead of allowing contracts to be ts since it seems the pattern is to just use this library, but this is what the path seems to look like:

*   [x] Switch near-sdk-js to typescript (#82)
*   [x] Add support into JS CLI to build typescript contracts? (this one is a bit unclear because there is some convoluted logic in there)
*   [ ] Switch some or all examples to TS

Options for adding support to SDK CLI to build typescript contracts intuitively to me are:

*   Conditionally use the [typescript plugin for rollup](https://www.npmjs.com/package/@rollup/plugin-typescript) to be able to compile ts contract file 
      *   This would require checking the input file name for the type to be able to include typescript plugin or not – maybe both `index.js` and `index.ts` are checked for existence? Otherwise, someone doing ts contracts will just have to specify the `source` argument to the ts file.
*   Modify CLI to compile the TS contract file to ESM JS and use that file in the rollup
*   Don't change any tooling, developers can compile ts to JS themselves and just update the `source` argument to this file

ailisp:
To be clear, SDK can be written with TS (or JS + typedefs), but from the original requirement, we must not limit user to write contract with TypeScript (not near-sdk-ts)

austinabell:
> To be clear, SDK can be written with TS (or JS + typedefs), but from the original requirement, we must not limit user to write contract with TypeScript (not near-sdk-ts)

Yes, but having the option to write a contract in typescript is what is being referred to here. Thanks for making that point clear

ailisp:
> Add support into JS CLI to build typescript contracts

is added in #134 

> Switch some or all examples to TS

Had a first example in TS in #134. More examples are discussed offline and we plan to create issue per example to track

PR #82, opened 2022-06-03
feat: Switch SDK to TypeScript

austinabell:
Still trying to resolve some issues I've found along the way, and `near-bindgen.js` is particularly prickly because I don't know why a variable is assigned to a function then the `_get` function is added as a property. Feels strange

Somewhat yoinked the NAJ tsconfig but had to switch the target to es2020 since rollup wasn't acknowledging the exports with commonjs


ailisp:
> Still trying to resolve some issues I've found along the way, and near-bindgen.js is particularly prickly because I don't know why a variable is assigned to a function then the _get function is added as a property. Feels strange

@itegulov was giving the same comment and I agree with you: https://github.com/near/near-sdk-js/pull/21#discussion_r873482329

But I cannot get a more elegant version work, do you have some ideas?

austinabell:
Okay, this PR is about at the point where I'll push it to. I left a few files as JS because I just don't have much context on what they are, namely `near-bindgen-exporter.js` and `near-bindgen.js` (the latter can't be converted to TS right now it seems).

When reviewing, please check the types carefully, we might not catch some issues with the APIs until we implement support for building typescript contracts (second point of #79). I made some assumptions and tried to infer based on the jsvm code, but it's extremely possible I missed some nuance.

The opinionated part about these APIs is if interfaces that take `string`, which refers to bytes, should also allow `Uint8Array`. Unclear if this is supported by jsvm, and if not, should there be support added or the conversion happening internally? cc @ailisp

ailisp:
> The opinionated part about these APIs is if interfaces that take string, which refers to bytes, should also allow Uint8Array. 

Makes sense. We recently add this API: https://github.com/near/near-sdk-js/commit/dc4403c7a099c052d454c79afb5965f1c1234e79#diff-3274f1a37032fb0ae4e2823def0007c634e869ae0dfc304ff6a12c36513c3a52

We require user to pass a bytes wrapped with the `bytes()` function, which typechecks the string and Uint8Array and converts Uint8Array. I slightly prefer this option over implicitly convert Uint8Array every API, for performance reason

> Unclear if this is supported by jsvm, and if not, should there be support added or the conversion happening internally? cc @ailisp


It is not. quickjs doesn't expose capabilities to manipulate uint8array to C API. Add support is a good approach, we have already extend quickjs C APIs in a few ways. In long term, it's better than the `bytes` API or other forms of JS side conversion in every function in `api.js`.

For the scope of this PR, I think it's best just to rely on the `bytes` API

lib/api.d.ts
```diff
@@ -0,0 +1,40 @@
...
+export declare function predecessorAccountId(): string;
+export declare function blockIndex(): BigInt;
+export declare function blockHeight(): BigInt;
+export declare function blockTimestamp(): BigInt;
+export declare function epochHeight(): BigInt;
+export declare function attachedDeposit(): BigInt;
+export declare function prepaidGas(): BigInt;
+export declare function usedGas(): BigInt;
+export declare function randomSeed(): string;
+export declare function sha256(value: string): string;
```
ailisp, line 13: I suggest to typedef string to bytes, then set those API taking bytes with that interface, such as sha256, storageWrite, logUtf8 etc. Some APIs are takes/returns string, such as jsvmMethodName. Could you use reference this for which ones are bytes: https://github.com/near/near-sdk-js#near-sdk-js-api-reference

src/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+  let ret = "";
+  for (let e of array) {
+    ret += String.fromCharCode(e);
+  }
+  return ret;
+}
+
+// TODO this function is a bit broken and the type can't be string
+// TODO for more info: https://github.com/near/near-sdk-js/issues/78
+export function bytesToU8Array(bytes: any): Uint8Array {
```
ailisp, line 11: Also here, typedef `Bytes` and make it `bytesToU8Array(bytes: Bytes): Uint8Array` also address this, Considering it a type with semantic restrictions, string just happens to be the internal representation. So it becomes clear a string can't be directly pass to it.

src/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+  let ret = new Uint8Array(bytes.length);
+  for (let i in bytes) {
+    ret[i] = bytes.charCodeAt(i);
+  }
+  return ret;
+}
+
+export function bytes(strOrU8Array: string | Uint8Array): string {
```
ailisp, line 19: And this would be `export function bytes(strOrU8Array: string | Uint8Array): Bytes {`

ailisp (commented):
Nice job! Just one point I think explicitly typing Bytes type to differentiate it with string would be great.

lib/api.d.ts
austinabell: I suppose it's fine to alias this for now. Just feels weird that there are these semantics that is specific to this project and not expressed in some way through the type system to avoid misuse

src/utils.ts
austinabell: This current functionality doesn't work. You can't iterate over the string as you're doing here with the for loop. Must be some unstable part of JS that this is functional.

Do you want me to switch the functionality to one that's consistent with `string`? This might be a breaking change -- which I was trying to avoid

src/utils.ts
```diff
@@ -0,0 +1,42 @@
+// Type alias of bytes as string. This is done because internally, JSVM accepts `string`
+// as bytes. In the future, this can be updated to a separate type if added support
+// within quickjs/jsvm.
+export type Bytes = string;
+
+export function u8ArrayToBytes(array: Uint8Array) {
```
ailisp, line 6:
```suggestion
export function u8ArrayToBytes(array: Uint8Array): Bytes {
```

src/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+  let ret = "";
+  for (let e of array) {
+    ret += String.fromCharCode(e);
+  }
+  return ret;
+}
+
+// TODO this function is a bit broken and the type can't be string
+// TODO for more info: https://github.com/near/near-sdk-js/issues/78
+export function bytesToU8Array(bytes: any): Uint8Array {
```
ailisp, line 16:
```suggestion
export function bytesToU8Array(bytes: Bytes): Uint8Array {
```

src/utils.ts
austinabell: Yeah, that's the `String` object, not the `string` primitive. It works, but TS doesn't like it. Is it acceptable to change the code to a traditional for loop?

src/utils.ts
austinabell: https://github.com/near/near-sdk-js/pull/82#discussion_r902096369

src/utils.ts
ailisp: Ah got you! I didn't know TS doesn't like loop on the string primitive. 
>  Is it acceptable to change the code to a traditional for loop?

Yes

ailisp (approved):
Awesome! Thank you

ailisp:
@volovyk-s should we merge this to `develop` branch now?

package.json
```diff
@@ -29,17 +32,23 @@
...
     "@rollup/plugin-node-resolve": "^13.1.1",
+    "rollup": "^2.61.1",
+    "rollup-plugin-sourcemaps": "^0.6.3",
     "yargs": "^17.5.1"
   },
   "files": [
     "cli",
     "res",
     "scripts",
     "src"
```
volovyk-s, line 46: Now we need to package `lib` folder instead of `src`.

src/collections/lookup-map.ts
```diff
@@ -1,37 +1,40 @@
...
 export class LookupMap {
-    constructor(keyPrefix) {
+    readonly keyPrefix: Bytes;
+
+    constructor(keyPrefix: Bytes) {
         this.keyPrefix = keyPrefix
     }
 
-    containsKey(key) {
+    containsKey(key: Bytes): boolean {
```
volovyk-s, line 11: What is the UX here? Can user just pass a string?

volovyk-s (commented):
Right now master is frozen and we are continuing development in `develop`. I think we should merge this PR there and not diverge branches.

Also, let's add or rewrite at least one contract example written in TS. Maybe we will need to update build scripts in order to make it work.

Other comments:
- let's not update `res/` build artifacts. (if we do not have a reason to do that).

austinabell:
> Right now master is frozen and we are continuing development in `develop`. I think we should merge this PR there and not diverge branches.

This should probably be the default branch or have been mentioned before. Will be very frictional for external developers to have to rebase and handle merge conflicts from the default branch of master. I'll fix now

> 
> Also, let's add or rewrite at least one contract example written in TS. Maybe we will need to update build scripts in order to make it work.

This is a separate issue, that can be tackled in a separate PR. #79 mentions this. Want to keep the scope of this PR small

> Other comments:
> 
> * let's not update `res/` build artifacts. (if we do not have a reason to do that).

Yep, wasn't sure if things were stale and not updated. Seems these do not have a canonical build, either, which might want to be looked into at some point.

volovyk-s:
@austinabell agree it is our fault :/ We made this decision last week.

Ok, let's merge this code to `develop`, we can address other issues in separate PRs.

@ailisp @AustinBaggio I can not change the default branch here. Can you do it?

austinabell:
sorry, had to squash changes because rebasing from master target to develop pulled some extra commits and rebasing while trying to exclude the jsvm/qjs changes was a pain.

Everything was the same as before, except I will point out that when updating to develop [I changed the interface slightly for promiseResult](https://github.com/near/near-sdk-js/pull/82/files#diff-769911c416ccf8514d8fd941ae0abe8fb5c606ade0c218e22151a5f5f9f3d700R333-R342) since it didn't make much sense to have as the way it was. Happy to change back and define the interfaces for this if you'd like.

src/collections/lookup-map.ts
austinabell: Opened https://github.com/near/near-sdk-js/issues/117 to track this

austinabell:
`develop` branch is borked, seems like since the last CI run (13h ago). Don't have context to what has changed but didn't seem to be a non-deterministic issue

ailisp:
I also don't have permission to change the default branch to `develop`. I think we need a github admin @AustinBaggio do you have the permission, if not, we can ask Jim then Max

austinabell:
> Your new one is more succinct, but one naming is confusing - NotReady is not a kind of failure. Rust sdk's naming PromiseResult sounds more accurate. If choosing Result, it should also include the success case - I feel the previous one is closer to the Rust SDK's 

I agree, I wasn't very happy with the naming. The goal was to express "either bytes are ready and success, or handle other cases"

> I also don't have permission to change the default branch to `develop`. I think we need a github admin @AustinBaggio do you have the permission, if not, we can ask Jim then Max

Very unlikely Baggio has admin. @near-arthur has admin I believe, though

volovyk-s (approved):
Great contribution @austinabell !
P.S. I have fixed 1 test by increasing gas. Not sure what the reason is, but if you have ideas, let's fix that in a separate PR.

PR #96, opened 2022-06-06
Update README.md to describe the bytes type

ailisp:
Differentiate it from String, to be precise on typing of APIs


volovyk-s (approved):
Ok! In JS it is not even a breaking change, hah.

PR #103, opened 2022-06-09
Add explicit bytes to the interface and test low level APIs on bytes

ailisp:
More tests to test panic, panic_utf8, read write state to be added


tests/__tests__/bytes.ava.js
```diff
@@ -0,0 +1,211 @@
...
+});
+
+test('Log unexpected types not logging', async t => {
+    const { jsvm, ali, testContract } = t.context.accounts;
+
+    let r = await ali.callRaw(jsvm, 'call_js_contract', encodeCall(testContract.accountId, 'log_unexpected_input_tests', ''), { attachedDeposit: '100000000000000000000000' });
+    t.deepEqual(
+        r.result.receipts_outcome[0].outcome.logs,
+        [
+            "",
```
volovyk-s, line 67: Should we pt some explanatory message here? Something similar to `String encoding is bad UTF-8 sequence.`

volovyk-s (approved):
Looks cool!

src/api.js
```diff
@@ -82,7 +82,11 @@ export function ecrecover(hash, sign, v, malleabilityFlag) {
 }
 
 export function panic(msg) {
-    env.panic(msg)
+    if (msg !== undefined) {
```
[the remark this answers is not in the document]
ailisp: panicUtf8 / panicUtf16's interface is `panicUtf8(msg: bytes)`, and panic is `panic(msg?: any)`. So `panicUtf8/16` user must call with a msg in bytes, otherwise the behavior is undefined/not guaranteed.

For panic, user can pass any kind of object or nothing. If nothing is passed, quickjs will set msg as `undefined` (same behavior as nodejs), in this case we want to invoke host function `panic()`, instead of `panic(undefined)`

tests/__tests__/bytes.ava.js
```diff
@@ -0,0 +1,225 @@
...
+    r = await ali.callRaw(jsvm, 'call_js_contract', encodeCall(testContract.accountId, 'panic_js_number', ''));
+    t.deepEqual(
+        r.result.receipts_outcome[0].outcome.status.Failure.ActionError.kind.FunctionCallError.ExecutionError,
+        'Smart contract panicked: 356'
+    );
+
+    r = await ali.callRaw(jsvm, 'call_js_contract', encodeCall(testContract.accountId, 'panic_js_undefined', ''));
+    t.deepEqual(
+        r.result.receipts_outcome[0].outcome.status.Failure.ActionError.kind.FunctionCallError.ExecutionError,
+        'Smart contract panicked: explicit guest panic'
```
ailisp, line 199: There's no way to tell if a argument is not passed, or it's passed but undefined, same as nodejs:
```
Welcome to Node.js v18.2.0.
Type ".help" for more information.
> function a(x) { console.log(x) }
undefined
> a(undefined)
undefined
undefined
> a()
undefined
undefined
```

PR #106, opened 2022-06-16
Standalone merge

volovyk-s:
- All standalone related files moved to this branch
- Build script refactored
- Standalone build is using the `<OS>-<ARCH>-qjsc` file from the `res` folder
- `jsvm/vendor` folder is deleted from  `.gitignore` because it is used during the build process (we should find other way to do it or add only the necessary files)


ailisp:
You did it so fast! Nice job!

> we should find other way to do it or add only the necessary files

The binaryen can be dropped, just include the wasi-stub binary (one for each operating system).

The major pain is we shipped a setup script, which downloads c compiler. It has not only a binary, but many, and many lib, headers, etc.

I think use your approach to try shipping the compiler package, we need to ship binaries for Mac/M1 Mac too. Shipping the compiler package or use the setup script approach has different pros and cons.

The other way I can think of is using a dockerfile, but it isn't perfect too.

ailisp:
Think for a while, IMO it's still better to keep `vendor/` in gitignore. Because we're unlikely to make any changes to it, so obtained it via `setup.sh` as a bundle instead of part of source would be better.

ailisp:
I also move the standalone readme to top level, the original readme to jsvm, since standalone is going to be primary. Current build success and test passing without `stubs` folder, after the wasi-stub tool was introduced, so that is removed. Also a bit reorganize: setup.sh and vendor to toplevel because they are for both jsvm and standalone.

volovyk-s:
Let's use develop for PRs (cloned from current master). It will allow us to keep a stable version in master and move fast in develop.

volovyk-s:
@ailisp feel free to merge it to `develop` if it's blocking you.
The next step is to port `.sh` script to our cli. I can do that here or in a separate PR.

ailisp (approved):
Great idea on develop branch!

PR #114, opened 2022-06-21
Build script

volovyk-s:
`near-sdk src/index.js` -> generates contract.wasm
`near-sdk src/index.js build/contract.base64` -> generates contract.base64
`near-sdk src/index.js build/contract.wasm` -> generates contract.wasm

The build process still depends on QuickJS and the `vendor` folder with other dependencies. We will need to figure out a strategy for installing/using deps.

We will need to rewrite JS code generation in order to finalize the build process and be able to refactor examples.


cli/cli.js
```diff
@@ -81,24 +96,60 @@ async function createJsFileWithRullup(sourceFileWithPath, rollupTarget) {
 }
 
 async function cratreHeaderFileWithQjsc(rollupTarget, qjscTarget) {
```
ailisp, line 98: typo: createHeaderFileWithQjsc

ailisp:
> We will need to rewrite JS code generation in order to finalize the build process and be able to refactor examples.

Is this achieved in https://github.com/near/near-sdk-js/pull/114/files#diff-54a0aa70bbfc306c7e92f0c21a52bbd6b76f53dd70e4d87714530195e58bab79R117 ?

> The build process still depends on QuickJS and the vendor folder with other dependencies. We will need to figure out a strategy for installing/using deps.

Can be consider as part of DPT-1141. For the build script into sdk cli, this PR is great

volovyk-s:
> Is this achieved in https://github.com/near/near-sdk-js/pull/114/files#diff-54a0aa70bbfc306c7e92f0c21a52bbd6b76f53dd70e4d87714530195e58bab79R117 ?

No, this part should be good. I'm talking about decorators. Now, the contract can be deployed, but it fails with JSVM storage errors. I believe we have a separate task for that.

> Can be consider as part of DPT-1141. For the build script into sdk cli, this PR is great

Yes!

ailisp:
> No, this part should be good. I'm talking about decorators. Now, the contract can be deployed, but it fails with JSVM storage errors. I believe we have a separate task for that.

I see, That is closer to success than expected!

PR #115, opened 2022-06-21
Standalone api

ailisp:
Add standalone API to api.js; For now, before two SDKs are unified, move jsvm examples and tests to `jsvm/`.


volovyk-s (approved):
Cool. I think we will need to separate standalone and enclaved APIs in the future. Or delete the enclaved one.

ailisp:
> Cool. I think we will need to separate standalone and enclaved APIs in the future.

Yeah. It is more challenging than I thought to configure rollup, I created DPT-1201 on jira

PR #119, opened 2022-06-23
fix: examples package.json and update lockfile

austinabell:
Noticed these were broken on develop, seems like the ava tests fail, though. I think it might have to do with the workspaces-js update? cc @volovyk-s 

Figured I'd make this fix in a separate PR and to ensure the CI is broken on this branch


ailisp:
Is the test passing locally? It's possible that too many `beforeEach` tries to launch sandbox hit some resource limitation on CI.

volovyk-s:
@austinabell I saw this failure as well. We do not need that script, since it is just a part of `build:non-fungible-token`.

volovyk-s (approved):

PR #122, opened 2022-06-24
Adapt decorators, collections, near-contract for standalone design

volovyk-s:
I have ported everything needed for the standalone design. The question is: do we want to keep enclaved files, examples etc? Now they are broken.
In order to support them, we will ned to have a separate version of collections, decorator, near-contract + separate folder with examples and tests (the last part is already in the project).
My saggestion is to clean the project from JSVM completely.


ailisp:
> In order to support them, we will ned to have a separate version of collections, decorator, near-contract + separate folder with examples and tests (the last part is already in the project).
My saggestion is to clean the project from JSVM completely.

Yeah, this is really annoying. Unless we can quickly figure out a way to conditional import/compile, then drop jsvm completely is the reasonable way

ailisp (approved):
PR looks good!

volovyk-s:
@ailisp ok, I will delete the enclaved CI for now. Let's deal with refactoring later and only if we will decide to have both designs.

PR #123, opened 2022-06-24
Standalone examples

volovyk-s:
- Moved all examples to standalone
- Moved and rewritten all tests to standalone

Generally speaking, contracts are unchanged. The only thing that we need to fix is cross-contract calls (replace `near.jsvmCall`). I would prefer to do that in a separate PR.


ailisp (approved):
Excellent! It's extremely helpful that all examples except cross contract call works!

PR #126, opened 2022-06-27
Storage api test

ailisp:
Sorry commit is a bit messed, should be okay by squash


log
```diff
@@ -0,0 +1,1339 @@
+i panic_utf8
```
volovyk-s, line 1: Probably this file is added by accident.

volovyk-s (approved):

PR #128, opened 2022-06-28
Promise API examples

volovyk-s:
- NFT and cross-contract-call examples rewritten for standalone
- near.log() API improved. Now we can pass several parameters, objects, undefined, null, etc. Everything will be converted to a string and logged.
Examples: 
image


lib/api.js
```diff
@@ -257,15 +259,15 @@ export var PromiseResult;
 })(PromiseResult || (PromiseResult = {}));
 export function promiseResult(resultIdx) {
     let status = env.promise_result(resultIdx, 0);
-    if (status === PromiseResult.Successful) {
+    if (status == PromiseResult.Successful) {
```
volovyk-s, line 262: It was throwing "Unexpected return code.."

ailisp: Ah, the problem is env.promise_result returns a big int (`1n`), but `PromiseResult.Successful = 1`

examples/src/cross-contract-call.js
```diff
@@ -10,7 +10,18 @@ class OnCall extends NearContract {
...
+        near.promiseThen(promise, near.currentAccountId(), '_set_person_on_call_private', bytes(JSON.stringify({ accountId: accountId })), 0, 30000000000000);
+    }
+
+    @call
+    _set_person_on_call_private({ accountId }) {
+        near.log(`_set_person_on_call_private called, accountId ${accountId}`)
+        if (near.currentAccountId() == !near.predecessorAccountId()) {
+            near.panic('Function can be used as a callback only')
+        }
+        const status = near.promiseResult(0).replaceAll('\"', ''); // TODO: Any ideas why we have "AVAILABLE" instead of AVALABLE?
```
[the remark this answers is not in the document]
ailisp: The problem is in high level contracts (written with `@nearbindgen` and `NearContract`. In this case, `status-message` contract), the return value is auto-serialized:
```
  static serializeReturn(ret: any) {
    return JSON.stringify(ret);
  }
```
And `JSON.stringify('AVAILABLE') -> '"AVAILABLE"'` 

So we can `JSON.parse` to deserialize the result back to string `AVAILABLE`.

src/api.ts
```diff
@@ -11,6 +11,8 @@ interface Env {
 // env object is injected by JSVM
 declare let env: Env;
 
+// TODO: rewrite api.js to named parameters?
```
ailisp, line 14: IMO; current interface is okay, as we can indicate the order of arguments match https://nomicon.io/Proposals/system-methods.

Most APIs, except for promise APIs, have a short param list. For promise APIs, these many params are painful. I think for these APIs, we can consider rust sdk's high level promise API, see whether we can build it in js sdk (Maybe in 2.0)

src/api.ts
```diff
@@ -11,6 +11,8 @@ interface Env {
...
+// TODO: should have the same interface as console.log
```
ailisp, line 15: Good catch! `log` can actually works on `message: any` instead of `string`: https://github.com/near/near-sdk-js/blob/cb152ebb16e4318207cc0a1b0507b6e6ecbeb3da/tests/src/log_panic_api.js#L9

So we can just modify the interface here

src/api.ts
volovyk-s: Agree. And this API should never change, let's keep it as it is for now.

src/api.ts
ailisp: I see! Your enhanced `near.log` is awesome!

examples/src/cross-contract-call.js
```diff
@@ -10,7 +10,18 @@ class OnCall extends NearContract {
...
-        const status = near.jsvmCall('status-message.test.near', 'get_status', { account_id: accountId })
+        const promise = near.promiseBatchCreate('statusmessage.test.near')
+        near.promiseBatchActionFunctionCall(promise, 'get_status', bytes(JSON.stringify({ account_id: accountId })), 0, 30000000000000)
+        near.promiseThen(promise, near.currentAccountId(), '_set_person_on_call_private', bytes(JSON.stringify({ accountId: accountId })), 0, 30000000000000);
+    }
+
+    @call
+    _set_person_on_call_private({ accountId }) {
+        near.log(`_set_person_on_call_private called, accountId ${accountId}`)
+        if (near.currentAccountId() == !near.predecessorAccountId()) {
```
ailisp, line 21: maybe typo?
```suggestion
        if (near.currentAccountId() !== near.predecessorAccountId()) {
```

ailisp (approved):
Great enhancement and examples!

examples/src/cross-contract-call.js
volovyk-s: My VSCode formatting works this way :/ But yes, I do not like it too.

ailisp: Looks like the VSCode formatting bug! `x == !y` is interpreted as `x == (!y)`, and it seems not what we want:
```
> 'a' == !'b'
false
> 'a' !== 'b'
true
```

volovyk-s: Wow, I was expecting the same behavior. Nice catch!

PR #133, opened 2022-07-12
Standalone Dependencies

volovyk-s:
All Cmake logic deleted/moved to separate repos.
- `binaryen`/`wasi-stub` built in a separate repo, released artifacts downloaded in `post-install` 
- `QuickJS` moved to a fork, released artifacts downloaded in `post-install`
- `wasi-sdk` artifacts downloaded in `post-install`
 
Other changes:
- `near-bindgen-exporter` added to `lib`, since `src` is not in the npm package


cli/post-install.js
```diff
@@ -0,0 +1,66 @@
+import { executeCommand } from './utils.js';
+
+const OS = await executeCommand('uname -s', true);
+const ARCH = await executeCommand('uname -m', true);
+
+if (OS !== 'Linux' && OS !== 'Darwin') {
```
ailisp, line 6: I'm not very sure on this: windows doesn't have a uname command. If so, we want to try catch above commands to catch windows.

ailisp (approved):
This is absolutely beautiful work! Our sdk js finally becomes a clean js project

PR #134, opened 2022-07-13
TypeScript contract build support and counter example in TypeScript

ailisp:
- build works and test pass.
- Modify interface of decorators, reimplement `@NearBindgen` to work with both ts and js
- Fix warnings given by babel

Note: Due to we need code transformation by our babel plugins, there is a concern of whether we should put tsc in the build pipeline. The recommended way is [babel for transpiling and tsc for type checking](https://www.typescriptlang.org/docs/handbook/babel-with-typescript.html#babel-for-transpiling-tsc-for-types). It's also pretty challenging with the alternative: use tsc in the build pipeline together with babel: 
- if tsc is applied first, the `@NearBindgen` would be lost, and we would be unable to see it as marker and code generation on top of it. 
- If babel applied then tsc, it doesn't work, because babel require babel-typescript-preset to parse TS AST, and after that, it already transform ts to js

So I think we should proceed the recommended way.


examples/package.json
```diff
@@ -5,22 +5,25 @@
...
     "build:counter-lowlevel": "near-sdk build src/counter-lowlevel.js build/counter-lowlevel.wasm",
+    "build:counter-ts": "near-sdk build src/counter.ts build/counter-ts.wasm",
     "build:cross-contract-call": "near-sdk build src/status-message.js build/status-message.wasm && near-sdk build src/cross-contract-call.js build/cross-contract-call.wasm",
     "build:fungible-token-lockable": "near-sdk build src/fungible-token-lockable.js build/fungible-token-lockable.wasm",
     "build:fungible-token": "near-sdk build src/fungible-token.js build/fungible-token.wasm",
     "build:non-fungible-token": "near-sdk build src/non-fungible-token-receiver.js build/non-fungible-token-receiver.wasm && near-sdk build src/non-fungible-token.js build/non-fungible-token.wasm",
     "build:status-message-collections": "near-sdk build src/status-message-collections.js build/status-message-collections.wasm",
-    "test": "ava && yarn test:counter-lowlevel",
+    "test": "ava && yarn test:counter-lowlevel && yarn test:counter-ts && yarn test:ts",
+    "test:ts": "yarn run tsc",
```
volovyk-s, line 21: I guess this command will not execute ts tests.

volovyk-s: And I do not think we need a separate test command for TS-based scripts.

volovyk-s:
> Note: Due to we need code transformation by our babel plugins, there is a concern of whether we should put tsc in the build pipeline. The recommended way is [babel for transpiling and tsc for type checking](https://www.typescriptlang.org/docs/handbook/babel-with-typescript.html#babel-for-transpiling-tsc-for-types). It's also pretty challenging with the alternative: use tsc in the build pipeline together with babel:

So babel can do what is usually done by the `tsc` command?

examples/package.json
```diff
@@ -5,22 +5,25 @@
...
   "scripts": {
-    "build": "yarn build:clean-state && yarn build:counter && yarn build:counter-lowlevel && yarn build:cross-contract-call && yarn build:fungible-token-lockable && yarn build:fungible-token && yarn build:non-fungible-token && yarn build:status-message-collections && yarn build:status-message",
+    "build": "yarn build:clean-state && yarn build:counter && yarn build:counter-lowlevel && yarn build:counter-ts && yarn build:cross-contract-call && yarn build:fungible-token-lockable && yarn build:fungible-token && yarn build:non-fungible-token && yarn build:status-message-collections && yarn build:status-message",
     "rebuild": "cd .. && yarn build && cd examples && rm -rf node_modules && rm -rf build && yarn && yarn build",
     "build:status-message": "yarn near-sdk build src/status-message.js build/status-message.wasm",
     "build:clean-state": "near-sdk build src/clean-state.js build/clean-state.wasm",
     "build:counter": "near-sdk build src/counter.js build/counter.wasm",
     "build:counter-lowlevel": "near-sdk build src/counter-lowlevel.js build/counter-lowlevel.wasm",
+    "build:counter-ts": "near-sdk build src/counter.ts build/counter-ts.wasm",
```
volovyk-s, line 14: Looks like it is not added to CI (as well as the low-level counter example).

volovyk-s (approved):
Good work! Only minor questions.

examples/package.json
ailisp: Oh my naming is wrong. it's to type checking the code with tsc without building or testing. The result, is same as what is given by VSCode shows inline.

We need this because babel typescript plugin doesn't check errors. Maybe it's better to do it by `tsc --noEmit` as first step of `near-sdk-js build`, when typescript contract is given?

examples/package.json
ailisp: Fixed

examples/package.json
ailisp: > Maybe it's better to do it by tsc --noEmit as first step of near-sdk-js build, when typescript contract is given?

I did the simple fix for now. If you think ^ is better will do in next PR.

volovyk-s (approved):

PR #140, opened 2022-07-20
Add tsc to near-sdk build pipeline for tyescript contracts

ailisp:


volovyk-s (approved):
Nice! Is it stopping the build if TSC find some problem?

ailisp:
Good question! I checked and yes it is!

PR #143, opened 2022-07-21
Port all jsvm tests to standalone

ailisp:
I'll work on improve the collections, so first port those tests to ensure they are still working with the up coming improvement


tests/src/lookup-set.js
```diff
@@ -0,0 +1,41 @@
...
+        super()
+        this.lookupSet = new LookupSet('a');
+    }
+
+    deserialize() {
+        super.deserialize();
+        this.lookupSet = Object.assign(new LookupSet, this.lookupSet);
+    }
+
+    @view
```
willemneal, line 21: Wouldn't  any non-call be a view? Or are any method without a decorator considered private?

volovyk-s (approved):
Tests, I love tests :) A lot of tests.

PR #144, opened 2022-07-21
Rename binary "near-sdk" -> "near-sdk-js"

gagdiez:
I have renamed the binary from "near-sdk" to "near-sdk-js" since we have more than one "near-sdk" and it could be confusing for the users.

Note: This PR includes also two other minor changes:
    1. A minor change in cli.js to raise an exception early if the "target" is wrong
    2. A small fix to the "package.json" since it was not running the tests in my computer


cli/cli.js
```diff
@@ -48,7 +48,12 @@ async function build(argv) {
     const TARGET_DIR = path.dirname(argv.target);
     const TARGET_EXT = argv.target.split('.').pop();
     const TARGET_FILE_NAME = path.basename(argv.target, `.${TARGET_EXT}`);
-    const TARGET_TYPE = TARGET_EXT === 'wasm' ? 'STANDALONE' : TARGET_EXT === 'base64' ? 'ENCLAVED' : undefined;
+
+    if(!["wasm", "base64"].includes(TARGET_EXT)){
+        throw new Error(`Unsupported target ${TARGET_EXT}, make sure target ends with .wasm or .base64`);
+    } 
+
+    const TARGET_TYPE = TARGET_EXT === 'wasm' ? 'STANDALONE' : 'ENCLAVED';
```
gagdiez, line 56: I included these small changes to immediately stop the building if the target is not correct

ailisp (approved):
Looks good to me. I think near-sdk-js is a better name for cli, but I'll let @austinabell and @volovyk-s to confirm!

austinabell:
> Looks good to me. I think near-sdk-js is a better name for cli, but I'll let @austinabell and @volovyk-s to confirm!

Don’t have an opinion. I don't think it should conflict with any Rust tooling, esp since we don't have a near-sdk cli anyway, so this is opinionated.

gagdiez:
@austinabell @volovyk-s  I do believe that having a tool called "near-sdk" when it only refers to "near-sdk-js" can be confusing. For example, we currently have a website with that name (near-sdk.io), but it only presents documentation for Rust.

I think that it is a simple change that will help save confusion to devs, and will also help us down the line when more SDKs are created for other languages.

austinabell:
> @austinabell @volovyk-s I do believe that having a tool called "near-sdk" when it only refers to "near-sdk-js" can be confusing. For example, we currently have a website with that name (near-sdk.io), but it only presents documentation for Rust.

The plan was that to be for all SDKs, but the plan later was to merge the docs together, so was abandoned.

I don't disagree with the changes by the way, I'm just saying there won't be technical conflicts

volovyk-s (approved):

volovyk-s:
No strong opinion on naming here.
Throwing errors early is a great change.
@gagdiez let's fix the conflict and merge it!

Issue #91, opened 2022-06-03
collections requiring keys and values be serialized to strings


austinabell:
This is unintuitive for developers, since it's unclear strings exactly are required without any type indications, but also it's a very limiting API.

Ideally, for functions available, my suggestion (opinionated) is to have low-level functions such as `getRaw`, `pushRaw`, `replaceRaw` along with other core interfaces possibly under another class and then build these high-level APIs on top of that with some convenience such as JSON stringifying keys and values.

The main problem is `string` should not be the key and value type of these structures because it will be too far from the js std library equivalents

ailisp:
Same as #90, strings here are Bytes, and these are actualy `getRaw` `pushRaw`. But the real `get` and `push` etc. like rust-sdk is coupled with the borsh serialization format. We don't have borsh coupled with near-sdk-js and also json deserialize doesn't make sense to every key values, so it seems not clear how to provide a `get`/`push`. Since this, I decide to call `getRaw` etc. just `get`, and let users deal with collections encoding after that.

austinabell:
Why would it not make sense? Wouldn't you want support for non-string/"Bytes" keys?

I just think the API as-is will be misused by people trying to use the `Vector` as an array or the collection maps as a javascript object accepting any value parameter and running into bugs, but maybe we need input from community/other developers for this.

ailisp:
OK. I leave my opinion: arbitrary object seems not intuitive as keys. Because object, if it has reference to other data structure, are typically not used as hash table keys in programming languages.

Arbitrary object can be values, but have a good, extensible interface to support it requires a good design.

ailisp:
@BenKurrek encountered this issue, showing this can be quite common in complex and real world contracts. I still need a good way to implement the get on top of getRaw. For write, `push`/`set`/... can be roughly as `pushRaw/setRaw(JSON.stringify())` (plus some unicode to bytes decoding), but for read, it doesn't know which class is the object and cannot reconstruct the object

gagdiez:
what if the schema (as in https://github.com/near/borsh-js) is set on initialization, and stored in the state as part of the structure?

ailisp:
@gagdiez That's a good idea. However a migration pattern / standard need to be introduced and implemented. Because if user has borsh-js schema S, then make some changes to the schema S'. With schema-included format like JSON, User can write code to handle both version with ease. With borsh, I think we need a deploy-time pattern to:
1. read the old schema
[...]

Issue #89, opened 2022-06-03
Automatic `Object.assign` calls when deserializing collections


austinabell:
[https://github.com/near/near-sdk-js/blob/f03eb1a01ee1efbad18a06ab1173edb359e1afdd/examples/status-message/src/status-message-collections.js#L12-L16](https://github.com/near/near-sdk-js/blob/f03eb1a01ee1efbad18a06ab1173edb359e1afdd/examples/status-message/src/status-message-collections.js#L12-L16)

Is an example of the code I'm referring to. I'm wondering why this can't happen within the deserialize function, or if this is documented somewhere?

ailisp:
`super.deserialize` deserialize the state as JSON, the problem is it's an object, say `this.records.keys`, it's like `{"length":3","prefix":"a"}`. But we need it to be a `Vector` class. Manually this is via ` this.records.keys = Object.assign(new Vector, this.records.keys)`

@volovyk-s and I have discussed some way to make this automatic in deserialize, but it is not very easy and not implemented. If you have a good idea to implement that would be cool!

austinabell:
so then can't this issue be kept to track making this simpler and less error-prone for developers?

PR #147, opened 2022-07-26
Automatic serialize and deserialize object in collections

ailisp:
Among a few other alternatives (see below), this seems the most developer friendly version of API. See new added test in this PR about how it works. The cons: add about 10KB to contract due to the superserial dependency; need to add a line to deserialize method of the contract to make sure the serializer is properly setup.

Addressed #91 . This approach would also work for other collections, but before adding them let's review and discuss!

The alternatives:
- a manually reviver pattern (or write this pattern / code generation to make it happens automatically): https://stackoverflow.com/questions/14027168/how-to-restore-original-object-type-from-json. Some similar variants:
  - https://clarle.github.io/yui3/yui/docs/json/json-freeze-thaw.html
  - https://rebolon.medium.com/json-reviver-pattern-how-to-restore-object-from-json-into-real-entity-737b1a220450
  - https://riptutorial.com/javascript/example/1389/serializing-and-restoring-class-instances
- use a enhanced JSON serializer library, that has ability to recover class. This PR uses superserial. Alternatives: https://github.com/shaochuancs/esserializer


BenKurrek:
So with this, all JS contracts would be ~15 $NEAR more expensive than Rust. This isn't a ton and if you're on testnet, it doesn't matter. If you're on mainnet, you probably have the $NEAR since you're deploying a prod level contract.

src/collections/lookup-map.ts
```diff
@@ -1,23 +1,42 @@
...
     }
 
+    getObject(key: any): any {
+        let storageKey = this.keyPrefix + this.serializer.serialize(key)
+        let raw = near.storageRead(storageKey)
+        if (raw !== null) {
+            return this.serializer.deserialize(raw)
+        }
+        return null
+    }
```
austinabell, line 38: I don't personally see the reason for having both supported like this. The whole use case for JS SDK is that it is easier to use. I have a strong but low confidence opinion that the lookup map should be typed as `LookupMap<K, V>` where the keys and values would be serialized and have that as the base methods (`get`, `remove`, etc.). If the types are generic, you shouldn't need this superserial library, do you?

If the duplication is kept, I don't see why this wouldn't be swapped where the more primitive ones would be called something else, and the simple methods to use are the name like `get`

src/collections/lookup-map.ts
```diff
@@ -1,23 +1,42 @@
...
 
     containsKey(key: Bytes): boolean {
         let storageKey = this.keyPrefix + key
         return near.storageHasKey(storageKey)
     }
 
+    containsObjectKey(key: any): boolean {
+        let storageKey = this.keyPrefix + this.serializer.serialize(key)
+        return near.storageHasKey(storageKey)
+    }
```
austinabell, line 24: If you keep these methods as is, you should probably point the expanded function at the other to make it more maintainable and avoid duplication. Something like
```suggestion
    containsObjectKey(key: any): boolean {
        return this.containsKey(this.serializer.serialize(key))
    }
```

volovyk-s:
I can see why we have such a design here, but anyway:
- I think we should avoid using external libraries in the SDK unless it is absolutely necessary.
- Having two functions for `set`/`get` is not the best DevEx.

I agree with Austin that we should try to implement collections using generics. No guarantees that this is possible, but it is something developers are used to and at the first glance it looks like something that can work.

ailisp:
> So with this, all JS contracts would be ~15 $NEAR more expensive than Rust. This isn't a ton and if you're on testnet, it doesn't matter. If you're on mainnet, you probably have the $NEAR since you're deploying a prod level contract.

No, **without this** standalone JS contract was 5 $NEAR more expensive than Rust (500 KB), This PR adds 10KB, which is 0.1 $NEAR more

ailisp:
@austinabell @volovyk-s Your point makes sense, I struggle in find a good naming.

In rust-sdk, it's `get` (borsh deserialize) and `get_raw` (bytes). In js-sdk, originally we had `get` (bytes). This PR was intended to have this style API: keep `get`, make `get(JSON)Object`, `getBorshObject`
Alternatively:
- make a breaking change to rename `get` to `getRaw`, and make room for new `get` to json deserialize
- also `get` -> `getRaw`, and a get takes an optional argument `options?`. By default, `get` would use JSON deserialize and can be set to use other deserialization method.

ailisp:
> I think we should avoid using external libraries in the SDK unless it is absolutely necessary.

About this, I can try to cherry pick the logic as part of sdk

volovyk-s:
> * make a breaking change to rename `get` to `getRaw`, and make room for new `get` to json deserialize

I think we should not be afraid to make a breaking change here. Consistency between JS and RS versions is more important.

ailisp:
> > I think we should avoid using external libraries in the SDK unless it is absolutely necessary.
> 
> About this, I can try to cherry pick the logic as part of sdk

Updates: it seems not possible to reduce logic from superserial - it completely rewrite instead of reviving on JSON.stringify/parse. Need to copy almost all code from library is not better than just import the library :sweat_smile: . 

After look into superserial's code, it produce JSON for plain object, array and strings but not in all cases. A `TestUser` class of `name` and `age` property is serialized as: `TestUser{"name":"wan2land","age":20}`. It's like a superset of json that still human readable.

We don't have to json serialize the state. For most use case where developer just load/save to state through nearbindgen, it doesn't matter what the format is. So it would be nice to use a lightweight, succinct format, schema included and JS class preserving library to allow fully automatic serialization/deserialization. JSON isn't ideal because it doesn't preserve JS's class. Borsh isn't ideal since it requires schema definition (also recall that the borsh-js support is hacky @volovyk-s ). superserial is so far a reasonable solution, but if you know any more popular library satisfy these requirements would be great:
- lightweight: <= 10KB quickjs bytecode
- succinct: at most as lengthy as JSON, the shorter the better
- schema included: schema is serialized so user can deserialize without specify a spec. Like json. Unlike borsh
- preserve JS class: instances can be deserialized as the original class's instance, instead of object.

austinabell:
> @austinabell @volovyk-s Your point makes sense, I struggle in find a good naming.
> 
> 
> 
> In rust-sdk, it's `get` (borsh deserialize) and `get_raw` (bytes). In js-sdk, originally we had `get` (bytes). This PR was intended to have this style API: keep `get`, make `get(JSON)Object`, `getBorshObject`
> 
> Alternatively:
> 
> - make a breaking change to rename `get` to `getRaw`, and make room for new `get` to json deserialize
> 
> - also `get` -> `getRaw`, and a get takes an optional argument `options?`. By default, `get` would use JSON deserialize and can be set to use other deserialization method.
> 
> - Make `get` a full-capable API: when pass it a key in bytes, it prefix the key and storageRead. When pass it an object key, it auto serialize it, prefix and storageRead

Noting that the duplicate method names just exist with the older version of the SDK collections. There are only the valid types in the one we are migrating to.

ailisp:
Two known issues:
- (bug): for unicode string, superserial may keep it as is, but our collections cannot store `>255` char. A fix would be change how string is serialize and deserialize in superserial: https://github.com/denostack/superserial/blob/4669fa731a74c1f9b6b3b50be5b23cdfedc31d26/serialize.ts#L18

- (enhancement): maybe solve in #89, but if not, the serializer is in every collection and quite annoying to deserialize the serializer itself.

volovyk-s (approved):
Cool! Sending it to review for DevRel.

PR #149, opened 2022-07-27
add storageByteCost to api and assert to utils

ailisp:
storageByteCost is useful in many storage cost related contracts. Copied from rust sdk. assert is commonly used in some of examples. Deserve as part of api.


volovyk-s (approved):
Nice refactoring

PR #154, opened 2022-07-29
Lookup map naive implementation

volovyk-s:
Not working yet, has the same issues with serialization.
Using generics (LookupMap<V> for the sake of simplicity, ideally it should be LookupMap<K, V>)


volovyk-s:
Agreed with @ailisp to use generics approach from this PR combined with serialization library. JSON.serialize/parse is very limited and works only in basic cases.

PR #161, opened 2022-08-03
fix TypeScript import without .ts by config rollup

ailisp:
The root problem was: tsc will reject `import ./foo.ts`, only accept `import ./foo`, but rollup was only accepting `import ./foo.ts`. Serhii pointed out tsc cannot support with `.ts` import:  https://github.com/microsoft/TypeScript/issues/37582. So the other way is to configure rollup to accept without `.ts`, which is this PR.


volovyk-s (approved):
The fix is great, but I think we need to add a more real-life example of the ts import.
As an example, we can move Classes from -parking-lot example to a separate model.ts file.
What do you think? I can do that if you are ok with it.

PR #168, opened 2022-08-04
Serialization

volovyk-s:


ailisp (approved):

ailisp:
@volovyk-s great job! I fixed the two issues you mentioned.

src/collections/vector.ts
```diff
@@ -15,15 +14,13 @@ function indexToKey(prefix: Bytes, index: number): Bytes {
 
 /// An iterable implementation of vector that stores its content on the trie.
 /// Uses the following map: index -> element
-export class Vector<E> {
+export class Vector {
```
austinabell, line 17: Curious why the removal of the generic? Won't using `unknown` be limiting for developers? (Asking just to learn/understand, ignore if you'd like)

volovyk-s: Version with generics appeared to be buggy and resource hungry. Deserialization of the class object is not a trivial task, it is still an open question of how much magic we want to have, and how much should be done by the user, so he is in control of all the computations happening on the chain.
The current design is not final, this is something stable we were able to ship right now.

Issue #131, opened 2022-06-30
Delete near.panic()


volovyks:
`near.panic('Message')` is doing exactly the same as `throw new Error('message')`. Let's delete `near.panic('Message')`, mentions that this `env.panic()` API is not used and refactor examples/tests.

ailisp:
Good idea!

volovyk-s:
@ailisp 
nomicon.io says that panic does not accept any arguments. Only panic_utf8 does.
https://nomicon.io/RuntimeSpec/Components/BindingsSpec/MiscellaneousAPI
but in our code, both functions accept message parameters: 
```JS
export function panic(msg?: string): never {
  if (msg !== undefined) {
    env.panic(msg);
  } else {
    env.panic();
  }
}

export function panicUtf8(msg: string): never {
  env.panic_utf8(msg);
}
```

Can I delete both of them? Can we replace all edge cases with `throw new Error(message)`?

ailisp:
> Can I delete both of them? Can we handle all edge cases with throw new Error(message)?

For panic, it's covered by throw new Error and can be safely deleted. For panicUtf8, I think practically it's not useful, **can be dropped**, although in theory not all edge cases are handled. 

In detail, panicUtf8 takes a decoded bytes, nearcore encode it back to UTF-8 and prints. In panic it takes JS string, log as is. For example, utf8 decode `水` is `[0xe6, 0xb0, 0xb4]`. `near.panicUtf8(bytes('\xe6\xb0\xb4'))` prints `水`. `near.panic('水')` prints `水`, near.panic(bytes('\xe6\xb0\xb4')) prints 'æ°´' (char 0xe6, char 0xb0, char 0xb4). You can check more cases in test_log_panic_api.ava.js

PR #171, opened 2022-08-09
near.panic() deleted

volovyks:
In this PR we are enforcing the usage of `throw Error(msg)` over the `near.panic(msg)` as a more familiar syntax for JS developers.
Also, it appeared that `throw Error(msg)` also prints the line where this error happened, which is useful.


lib/api.js
```diff
@@ -67,14 +67,7 @@ export function ecrecover(hash, sig, v, malleabilityFlag) {
...
-        env.panic(msg);
-    }
-    else {
-        env.panic();
-    }
-}
+// NOTE: "env.panic(msg)" is not exported, use "throw Error(msg)" instead
 export function panicUtf8(msg) {
     env.panic_utf8(msg);
 }
```
austinabell, line 73: is it intentional to keep this? Or is this intended to be if someone did not want to include the debug info (line number, etc)?

ailisp: See our discussion here: https://github.com/near/near-sdk-js/issues/131#issuecomment-1208869788

ailisp (approved):
Looks good!

Issue #174, opened 2022-08-10
Can't use bigint literals or do arithmetic with es5 target


austinabell:
Currently, the CLI statically targets es5 for compiling contracts, but this makes certain things within the SDK hard to do like performing arithmetic on the values returned from the API (I don't know if another target fixes this) or using bigint literals (like `100n` to signify a big int).

I don't even know what the recommended way to perform a simple action like:

```ts
near.prepaidGas() - 5000000
```

The closest thing I could find would be:
```ts
Number(near.prepaidGas()) - 5000000
```

but this is absolutely horrible because on overflow it just cuts off bits and doesn't error. Casting the BigInt to `any` or casting the right operand to `BigInt` seems to throw an error

ailisp:
It seems that we should use es2020 for compiling target, as bigint is in es2020 and quickjs support that. I'll check typescript decorator still works under that setting.

PR #176, opened 2022-08-11
es2020 target in tsc to allow bigint literal

ailisp:
Address #174


volovyks (approved):
JGTM! Let's add a simple test with the BigInt operation.

PR #177, opened 2022-08-16
NFT standard

### LINKED ISSUE #230: Implement NFT standard polymorphic to rust-sdk's

ailisp:
This implements all traits, functionalities and tests as in https://github.com/near/near-sdk-rs/tree/master/near-contract-standards/src/non_fungible_token. 

Still missing doc strings, will be added tomorrow. But this has been so long as blocked by features, working on higher priorities, etc. I'd like to see some comments!


volovyks:
@ailisp have you tried to run rust tests on this resulting wasm file? (is it possible?)

ailisp:
Thanks for your review.

> @ailisp have you tried to run rust tests on this resulting wasm file? (is it possible?)

Yes, `__tests__/my-nft/*.ava.js` and "Contract standard / examples" on ci does that.


> can we organize imports without paths? Maybe similarly to how it's done with main lib and examples?

Will do!

> Have you tried to use it in the external project?

I haven't. Updating the /example-contracts test to depend on a lib version should work too, I'll make it that way. 

>  I wander if standards should live in src, then be built to /lib and examples will live in /examples, as well as tests in main /tests folder

This is a good question! In rust sdk, near-contract-standard is a separate crate. We probably want this too, as most non-standard contract doesn't need near-contract-standards. It is a challenge for me to figure out where should I organize lib, examples and tests in this case. Maybe we should turn this repo into a workspace. Maybe it should live in a separate repo. If in same repo, whether examples should be organized in root /examples or their own examples, etc. A couple of things to discuss. 

So in this PR, I want only ensure it works as an external library, and let's have a discussion next week on ^ and make a separate PR to properly organize near-contract-standards in general!

near-contract-standards/src/non_fungible_token/events.ts
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * <https://github.com/near/NEPs/blob/69f76c6c78c2ebf05d856347c9c98ae48ad84ebd/specs/Standards/NonFungibleToken/Event.md>
+ *
+ * This is an extension of the events format (nep-297):
+ * <https://github.com/near/NEPs/blob/master/specs/Standards/EventsFormat.md>
+ *
+ * The three events in this standard are [`NftMint`], [`NftTransfer`], and [`NftBurn`].
+ *
+ * These events can be logged by calling `.emit()` on them if a single event, or calling
+ * [`NftMint::emit_many`], [`NftTransfer::emit_many`],
```
volovyks, line 13: Rust style calls here

near-contract-standards/src/event.ts
```diff
@@ -0,0 +1,19 @@
+import { near } from "near-sdk-js";
+
+export class NearEvent {
+  internal_to_json_string(): string {
```
volovyks, line 4: Should it be private? Looks like it is not intended to be executed by the user.

near-contract-standards/src/event.ts
```diff
@@ -0,0 +1,19 @@
+import { near } from "near-sdk-js";
+
+export class NearEvent {
```
volovyks, line 3: Is this class intended to be instantiated? Should it be abstract?

near-contract-standards/lib/non_fungible_token/utils.js
```diff
@@ -0,0 +1,42 @@
+import { near, assert } from "near-sdk-js";
```
volovyks, line 1: Let's add that trick for the lib folder so it's not shown in git changes.

examples/src/standard-nft/my-nft.ts
```diff
@@ -0,0 +1,219 @@
+import { NonFungibleToken } from "near-contract-standards/lib";
```
volovyks, line 1: Looks like a lot of work for somebody to use a standard token. I wander if there is a way not to override most of the functions if they remain "standard". (just a discussion for the future)

volovyks (approved):
Tremendous effort @ailisp ! We can iterate on all the comments later, it can be merged now. The code is clean and well-documented, and the tests look robust.

near-contract-standards/src/non_fungible_token/utils.ts
```diff
@@ -0,0 +1,83 @@
+import { near, assert, Bytes } from "near-sdk-js";
+import { AccountId } from "near-sdk-js/lib/types";
+
+export function bytes_for_approved_account_id(account_id: string): number {
+  // The extra 4 bytes are coming from Borsh serialization to store the length of the string.
```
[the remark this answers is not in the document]
ailisp: Good catch! I realized this is not actually 4 + 8, as we are not using borsh at all, let me figure out a right number!

near-contract-standards/src/non_fungible_token/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+  return account_id.length + 4 + 8;
+}
+
+export function refund_approved_account_ids_iter(
+  account_id: string,
```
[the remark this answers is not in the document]
ailisp: Fixed every string -> string in this file

near-contract-standards/src/non_fungible_token/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+    .reduce((a, b) => a + b);
+  const promise_id = near.promiseBatchCreate(account_id);
+  near.promiseBatchActionTransfer(
+    promise_id,
+    BigInt(storage_released) * near.storageByteCost()
+  );
+  near.promiseReturn(promise_id);
+}
+
+export function refund_approved_account_ids(
```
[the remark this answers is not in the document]
ailisp: It's straightforward 1:1 mapping rust's utils. It's just a helper function, so IMO this version is fine.

near-contract-standards/src/non_fungible_token/metadata.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/** This spec can be treated like a version of the standard. */
+export const NFT_METADATA_SPEC = "nft-1.0.0";
+
+/** Metadata for the NFT contract itself. */
+export class NFTContractMetadata {
+  public spec: string; // required, essentially a version like "nft-1.0.0"
+  public name: string; // required, ex. "Mosaics"
+  public symbol: string; // required, ex. "MOSIAC"
+  public icon: Option<string>; // Data URL
```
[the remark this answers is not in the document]
ailisp: Good point! `?` is certainly more storage efficient, when it's not present it doesn't get serialized.

near-contract-standards/src/non_fungible_token/impl.ts
```diff
@@ -0,0 +1,713 @@
...
+function expect_approval<T>(option: Option<T>): T {
+  if (option === null) {
+    throw new Error("next_approval_by_id must be set for approval ext");
+  }
+  return option;
+}
+
+/** Implementation of the non-fungible token standard.
+ * Allows to include NEP-171 compatible token to any contract.
+ * There are next traits that any contract may implement:
```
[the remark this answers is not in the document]
ailisp: :))) Replace all traits to interface!

near-contract-standards/src/non_fungible_token/events.ts
ailisp: Fixed everywhere in this file

near-contract-standards/src/event.ts
ailisp: You're right. Making it private

near-contract-standards/src/event.ts
ailisp: Yes, it's never instantiated, making it an abstract class

near-contract-standards/lib/non_fungible_token/utils.js
ailisp: Done!

examples/src/standard-nft/my-nft.ts
ailisp: Yeah, this need more discussion. In rust it's done via a few macros: `impl_non_fungible_token_core` etc. I didn't see actively maintainted or commonly used JS macro / code generator libraries to integrate with our build pipeline. One way maybe making a "macro" with babel, like our `includeBytes`

near-contract-standards/src/non_fungible_token/utils.ts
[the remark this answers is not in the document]
ailisp: It's actually difficult to figure out the size to storage refund.
in rust, it's `key: string, value: u64`, in borsh size is `4 + string length + 8 (for uint 64)`. In js it's `key: string, value: bigint`, and it's json serialized, not able to count this way.

 I'll write a different logic (by counting diff in `near.storageUsed()`.

examples/src/standard-nft/my-nft.ts
waynenilsen: i completely agree with @volovyks but i think that leads to deeper discussions about the codegen / transpilation process and _why_ something simpler may not work. I think that the desirable outcome is something like this

in this repo (or some other repo) the standard exists and is implemented

```typescript
@NearBindgen({ requireInit: true })
export class NonFungibleToken {
...
}
```

later, someone wants to customize (such as myself) and add a function, called, say, deposit to extend the abilities of the NFT as such 

```typescript
@NearBindgen({ requireInit: true })
export class MyNonFungibleToken extends NonFungibleToken {
    // i get everything from NonFungibleToken by default but now i can extend
   @call({ payable: true})
    deposit(args: DepositArgs): void { ... }
}
```

does that make sense?

we tried that but currently contracts can't be composed in this way, I don't have the errors at hand but I think a lot of it goes to the general codegen process related to the transpilation.

I can write a ticket to discuss further or we can chat in discord etc

ailisp:
@volovyks Addressed those easier ones. For more difficult ones, Will try to address them next Monday. As you said it's generally good - before the sprint planning I'll move unresolved / TBD points in separate issues to merge this PR :)

BenKurrek:
@volovyks @thisisjoshford From my understanding, if you pass in an array with the JS SDK, it uses positional arguments:
`nft_transfer([receiver_id, token_id, approval_id, memo])`.

This is not compatible with the standards and I believe we should use named arguments such as:
`nft_transfer({receiver_id: string, token_id: string, approval_id: number, memo: string})`.

By using positional arguments, it might not be compatible with some implementations and could break things. Please correct me if I'm wrong.

ailisp:
Pierre Le Guen has reached me and we agree to move to
`nft_transfer({receiver_id: string, token_id: string, approval_id: number, memo: string}).` format of argument. I think this is the issue mentioned by @thisisjoshford and @BenKurrek . Let me know if there's other issues!

examples/src/standard-nft/my-nft.ts
volovyks: Yes, I have something like this in mind as well. @waynenilsen issues with described problematics would be great!

ailisp:
Most comments are addressed. For three more challenging one, I move and track them in #288, #289, #290

examples/src/standard-nft/my-nft.ts
ailisp: I merged this one and created #290 . @waynenilsen feel free to experiment and may be create some proof of concept from latest `develop`!

PR #179, opened 2022-08-17
make UnorderedMap and Set helper methods private

### LINKED ISSUE #83: Switch APIs of internal collection functions to indicate private
Not sure what the convention is for obfuscating the functions from developers, but functions like `serializeIndex`, `deserializeIndex` should be more clear that they shouldn't be used.

I don't think there is any reason why they need to be exposed so either moving them out of the class so they aren't exposed or prefixing with `_` or `internal` should probably be done.

ailisp:
Fix #83

### DIFF
--- a/src/collections/unordered-map.ts
+++ b/src/collections/unordered-map.ts
@@ -5,6 +5,24 @@ import { Vector, VectorIterator } from "./vector";
 const ERR_INCONSISTENT_STATE =
   "The collection is an inconsistent state. Did previous smart contract execution terminate unexpectedly?";
 
+function serializeIndex(index: number): Bytes {
+  let data = new Uint32Array([index]);
+  let array = new Uint8Array(data.buffer);
+  return u8ArrayToBytes(array);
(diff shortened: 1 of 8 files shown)


src/collections/unordered-map.ts
```diff
@@ -40,19 +40,19 @@ export class UnorderedMap {
...
     let array = bytesToU8Array(rawIndex);
     let data = new Uint32Array(array.buffer);
     return data[0];
   }
 
-  getIndexRaw(key: Bytes): Bytes {
+  private getIndexRaw(key: Bytes): Bytes {
     let indexLookup = this.keyIndexPrefix + JSON.stringify(key);
     let indexRaw = near.storageRead(indexLookup);
     return indexRaw;
```
austinabell, line 58: why are these associated functions to begin with? Can't these live outside of the structure to avoid misuse (when ignoring or not using TS types)?

volovyks (approved):

PR #180, opened 2022-08-17
Idiomatic js length for collections

### LINKED ISSUE #85: Remove `len` and `isEmpty` methods from collections
I can't seem to find the equivalents with JS libraries. For example `Array` exposes `length` field as a number, but then doesn't have either of those functions.

Removing would make it so that this doesn't have to be maintained if this ever wanted to be removed in the future.

ailisp:
Fix #85


lib/collections/unordered-set.d.ts
```diff
@@ -1,12 +1,12 @@
...
 import { Vector } from "./vector";
 export declare class UnorderedSet {
-    readonly length: number;
     readonly prefix: Bytes;
     readonly elementIndexPrefix: Bytes;
     readonly elements: Vector;
     constructor(prefix: Bytes);
-    len(): number;
+    get length(): number;
+    set length(_l: number);
```
austinabell, line 9: There shouldn't be a set method here. That would lead to bugs

ailisp: This is required for our `deserialize` to work, see this comment: https://github.com/near/near-sdk-js/pull/180/files#diff-f67afbdc9e6ba66abb7302662cfab6bd61080e3a7e7247973f3c87bf8ab36d20R33

austinabell: interesting, okay, my assumption was that this could be at least private. All I'm pointing out is that this might be a footgun since this is part of the exposed API to developers

ailisp: Good point! I try to add private to the setter, that still builds and will technically make it not exposed API

volovyks (approved):

ailisp:
@volovyks Thanks!

PR #181, opened 2022-08-18
add all rust sdk non-deprecated types

ailisp:
in near-sdk-rs/src/types


src/types/account_id.ts
```diff
@@ -0,0 +1 @@
+export type AccountId = string;
```
volovyks, line 1: Should we move it to the primitives file?

volovyks (approved):
Great improvement!

PR #183, opened 2022-08-19
drop jsvm (aka enclave)

ailisp:
@thisisjoshford @BenKurrek From what we discussed in retro, please help make an announcement in blog / doc and maybe mention:
- Enclave, if there's a lot community interest, will be continue as a separate project
- There's an archived enclave, at `jsvm` branch, but note that it's feature behind standalone and not actively maintained


volovyks:
@ailisp I have deleted all JSVM-related APIs and logic.
+ I have deleted all mentions of "standalone name". I think we should not use it from now on. It is just NEAR JS SDK.

volovyks (approved):

PR #187, opened 2022-08-22
near-workspaces 3.2.1

volovyks:


ailisp (approved):
Nice update! We have macos test back!

PR #190, opened 2022-08-23
uint64 and uint128 overflow tests

ailisp:
Fixed the problem mentioned in https://github.com/near/near-sdk-js/discussions/182


volovyks (approved):
Nice! Not it is consistent.
Note: I have added arm64 QuickJS build to the v0.1.1 release so it can work on M1.

ailisp:
> Note: I have added arm64 QuickJS build to the v0.1.1 release so it can work on M1.

Great!

PR #192, opened 2022-08-23
ignore /lib in git diff

volovyks:
Works both locally and on GitHub


ailisp (approved):
Helpful setup! Unrelated to this PR: in ci can we check `lib/**` is built with the changes in `src/`

volovyks:
@ailisp 
in NAJ this problem was solved with a hook that is running `yarn build` before each commit. But it can be annoying because it makes commit process longer.

We should be able to run `yarn build` on the CI and then check the code with git diff, but we do not have such a setup in any of our libraries.

PR #198, opened 2022-08-27
fix(cli): paths

Olyno:
# Description

When using the CLI with different package managers, I was getting the error that the ``tsc`` command did not exist, when it was not.

So here is a pull request to fix these path problems.

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -2,6 +2,7 @@
 
 import fs from 'fs/promises'
 import path from 'path';
+import { fileURLToPath } from 'url';
 
 import yargs from 'yargs'
 import { hideBin } from 'yargs/helpers'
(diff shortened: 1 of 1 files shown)


volovyks:
Thanks, @Olyno! Can you please share what PM was problematic to use?
From what I can see in your PR, you suggest using global `tsc`. I do not think that we can assume that it is installed on every PC.

Olyno:
Hi @volovyks!

I got an error that ``tsc`` could not be found, which was not the case because it was installed locally. The problem is that, depending on the package manager used (yarn, pnpm or npm), the installation folder varies, but the command remains the same.

The best way would be to precede the command ``tsc`` with the package manager used, such as ``pnpm tsc``, in which case it would work perfectly. I made a quick fix, but as you mentioned, it is only available to those who have typescript installed globally (or yarn 2+ users).

Would you prefer that I use the package manager prefix?

volovyks:
> Would you prefer that I use the package manager prefix?
I think it will be a better solution. Now we have 0 prerequisites to use the SDK and it would be cool to keep it this way.

Olyno:
Hi @volovyks!
I close this pull request for the moment, because not relevant enough in my opinion. At the same time, I opened a pull request on the create-near-app repository (near/create-near-app#1901) which seems to me much more relevant at the moment, and which could add a real plus.

Don't hesitate to have a look at it if you have time!

volovyks:
Got it, thank you @Olyno !
Yes, we can come back to it. I believe that some parts of this PR can be merged.

PR #200, opened 2022-08-31
NearContract deletion, init, refactoring

volovyks:
- Init() is not required by default, now we have a `requireInit` flag in `NearBindgen`
- No need in `default()` function, deleted
- No need to extend `NearContract`. Class is deleted, and logic is moved to `NearBindgen`
- Added @initialize decorator for init functions. Function panics if called more then  once
- Now constructor is expected to be simplistic. No parameters, no side effects, only variable declarations


examples/src/counter.js
```diff
@@ -1,11 +1,10 @@
-import { NearContract, NearBindgen, near, call, view } from 'near-sdk-js'
+import { NearBindgen, near, call, view, initialize } from 'near-sdk-js'
 import { isUndefined } from 'lodash-es'
 
-@NearBindgen
-class Counter extends NearContract {
-    constructor({ initial = 0 }) {
-        super()
-        this.count = initial
+@NearBindgen({})
```
ailisp, line 4: Can this be written as `@NearBindgen()` or simply `@NearBindgen`?

examples/src/cross-contract-call.js
```diff
@@ -1,9 +1,14 @@
...
-@NearBindgen
-class OnCall extends NearContract {
-    constructor({ statusMessageContract }) {
-        super()
+@NearBindgen({ requireInit: true })
+class OnCall {
+    constructor() {
+        this.personOnCall = ''
+        this.statusMessageContract = ''
+    }
```
ailisp, line 8: Since `requireInit` , can we omit the constructor?

lib/build-tools/near-bindgen-exporter.js
```diff
@@ -4,7 +4,7 @@ export default function () {
         visitor: {
             ClassDeclaration(path) {
                 let classNode = path.node;
-                if (classNode.decorators && classNode.decorators[0].expression.name == 'NearBindgen') {
+                if (classNode.decorators && classNode.decorators[0].expression.callee.name == 'NearBindgen') {
```
ailisp, line 7: Not related to this PR, I thought we have set generated `.d.ts` and `.js` diff not showing in github, but seems it's not working

src/near-bindgen.ts
```diff
@@ -1,26 +1,56 @@
...
+            }
+
+            static _deserialize(value: string): Object {
+                return JSON.parse(value);
+            }
+
+            static _reconstruct(classObject: any, plainObject: JSON) {
+                for (const item in classObject) {
+                    if (classObject[item].constructor?.deserialize !== undefined) {
+                        classObject[item] = classObject[item].constructor.deserialize(plainObject[item])
```
ailisp, line 43: This is cool!!!

ailisp (approved):
Fantastic work! Solves at least three issues about initialization, default parameters. While most of mentioned behavior has been tested, I'd suggest to add two little tests about:

- init Function panics if called more then once
- the case when `requiredInit: false` but `@initialize` is provided

examples/src/counter.js
volovyks: Now `@NearBidgen` is defined as 
```JS
export function NearBindgen({ requireInit = false }: { requireInit?: boolean }) {
```
we are using named optional parameters.
It allows us to omit `requireInit` or other parameters and use the default value. Since it is a 'named' parameter, we can omit some of them. This design is extendable and this is why it needs to use a bit more complicated syntaxis.
 ```JS
@NearBindgen({})
```
Alternative 1: make all parameters required, NearBingen will always look like this:
```JS
@NearBindgen({ requireInit: true/false })
```

Alternative 2:  we can drop named parameters, but the design will not be intuitive:
```JS
@NearBindgen(true)
```

examples/src/cross-contract-call.js
volovyks: In this case, it is not possible to deserialize it since we do not know the list of properties.

lib/build-tools/near-bindgen-exporter.js
volovyks: Not sure... they were hidden for me.

examples/src/counter.js
ailisp: Got it! I think your implementation is better than the alternatives

examples/src/cross-contract-call.js
ailisp: I see!

examples/src/counter.js
miraclx: `@NearBindgen`, `@NearBindgen()` and `@NearBindgen({ requireInit: true })`

all work with this patch:

```patch
diff --git a/src/build-tools/near-bindgen-exporter.js b/src/build-tools/near-bindgen-exporter.js
index 8cd2780..6a34c9c 100644
--- a/src/build-tools/near-bindgen-exporter.js
+++ b/src/build-tools/near-bindgen-exporter.js
@@ -5,7 +5,7 @@ export default function () {
     visitor: {
       ClassDeclaration(path) {
         let classNode = path.node;
-        if (classNode.decorators && classNode.decorators[0].expression.callee.name == 'NearBindgen') {
+        if (classNode.decorators && (classNode.decorators[0].expression.name || classNode.decorators[0].expression.callee.name) == 'NearBindgen') {
           let classId = classNode.id;
           let contractMethods = {};
           for (let child of classNode.body.body) {
diff --git a/src/near-bindgen.ts b/src/near-bindgen.ts
index 9cf0d0e..73a175e 100644
--- a/src/near-bindgen.ts
+++ b/src/near-bindgen.ts
@@ -9,7 +9,9 @@ export function call(target: Object, key: string | symbol, descriptor: TypedProp
 export function view(target: Object, key: string | symbol, descriptor: TypedPropertyDescriptor<Function>): void {
 }

-export function NearBindgen({ requireInit = false }: { requireInit?: boolean }) {
+const DEFAULT_ARGS = { requireInit: false };
+
+export function NearBindgen({ requireInit = false }: { requireInit?: boolean } = DEFAULT_ARGS) {
     return <T extends { new(...args: any[]): {} }>(target: T) => {
         return class extends target {
             static _create() {
```

volovyks:
@ailisp tests added, merging.

PR #201, opened 2022-09-02
Add highlevel promise API

ailisp:
Add highlevel promise APIs as in Rust SDK


petarvujovic98:
Thanks for fixing the `BigInt` to `bigint` type definitions

volovyks (approved):
Clean design, I like it!

Issue #199, opened 2022-08-30
Adding instance type for vector, maps and sets


osalkanovic:
It would be beneficial to include instance type in vector, maps and sets. Something as `users = new Vector<User>().` Because overriding an instance with a type assertion ('as') is occasionally very painful

If you'd like, I can push a solution because I have a working example. 😄

volovyks:
Hey @osalkanovic ! We have made experiments in this direction, but serialization/deserialization appeared to be problematic. I would love to see your solution!

ailisp:
Implemented by @osalkanovic :tada: !

PR #203, opened 2022-09-02
Fix: maps, sets and vector types

osalkanovic:
Issue: https://github.com/near/near-sdk-js/issues/199


src/collections/lookup-map.ts
```diff
@@ -13,7 +13,7 @@ export class LookupMap {
         return near.storageHasKey(storageKey)
     }
 
-    get(key: Bytes): unknown | null {
+    get(key: Bytes): T | null {
```
ailisp, line 16: The issue is what this get is an object instead of `T`. For example, the CarSpec get will be `{id, color, price, engine}`, not `CarSpec{id, color, price, engine}`

src/collections/unordered-map.ts
```diff
@@ -59,21 +59,21 @@ export class UnorderedMap {
...
 
-  get(key: Bytes): unknown | null {
+  get(key: Bytes): T | null {
     let indexRaw = getIndexRaw(this.keyIndexPrefix, key);
     if (indexRaw) {
       let index = deserializeIndex(indexRaw);
       let value = this.values.get(index);
       if (value) {
-        return value;
+        return value as T;
```
ailisp, line 68: `as T` also doesn't solve the problem, it's only a ts sytax suger. When inspect the built JS contract, it is ignored. For example: https://www.typescriptlang.org/play?#code/LAKFGMBsEMGdYAQGFoCcDKAHApuBBvUBYhcAewDtYAXVAV3GrNQApM6AjSASzwugC22AFwIaqbhQDmASgIBfUIrAhI2aggAeCALwF+Q0QHJoR5Wo0BPXVoRxkaLLgDcESrDJqAdJDJSW1pI00BTg2GQAZg4YOOAyQA

ailisp (commented):
Thanks for the PR! There's still some problems on recovering the type of the object, so still cannot consider it as `T` :thinking:

src/collections/unordered-map.ts
petarvujovic98: @ailisp This could be solved by the new _brand_ feature of ES2022 ([here](https://inspiredwebdev.com/everything-new-in-es2022/#ergonomic-brand-checks-for-private-fields)). Using TS syntax here is better for DX since the developer won't look at the compiled code anyway.

src/collections/unordered-map.ts
```diff
@@ -128,32 +128,32 @@ export class UnorderedMap {
...
   }
 
   serialize(): string {
-    return JSON.stringify(this)
+    return JSON.stringify(this);
   }
 
   // converting plain object to class object
-  static deserialize(data: UnorderedMap): UnorderedMap {
+  static deserialize(data: UnorderedMap<unknown>): UnorderedMap<unknown> {
```
petarvujovic98, line 154: This should be a static function with a header something like this: 
```typescript
static deserialize<DataType>(data: UnorderedMap<DataType>): UnorderedMap<DataType> {
```

src/collections/unordered-set.ts
```diff
@@ -109,27 +110,27 @@ export class UnorderedSet {
...
   }
 
   serialize(): string {
-    return JSON.stringify(this)
+    return JSON.stringify(this);
   }
 
   // converting plain object to class object
-  static deserialize(data: UnorderedSet): UnorderedSet {
+  static deserialize(data: UnorderedSet<unknown>): UnorderedSet<unknown> {
```
petarvujovic98, line 124: Similar as [here](https://github.com/near/near-sdk-js/pull/203#r964436428)

src/collections/vector.ts
```diff
@@ -115,11 +115,11 @@ export class Vector {
   }
 
   serialize(): string {
-    return JSON.stringify(this)
+    return JSON.stringify(this);
   }
 
   // converting plain object to class object
-  static deserialize(data: Vector): Vector {
+  static deserialize(data: Vector<unknown>): Vector<unknown> {
```
petarvujovic98, line 122: Again, same as [above](https://github.com/near/near-sdk-js/pull/203#r964436428)

src/collections/unordered-map.ts
volovyks: Not sure what you mean. Why you are not using T here?

petarvujovic98: Right now this function is not generic amd will return an unknown. I am using `DataType` instead of `T` because that is just the generic name, you can use anything for `T` I just like using more descriptive names as it is self documenting for the library users

src/collections/unordered-map.ts
volovyks: Looks like this feature is not related to deserialization. Can you please elaborate on how it can help?

petarvujovic98: When constructing a collection we could set the private field (_brand_) to the type of the underlying data we store, this would be serialized and stored which means we could theoretically infer the type with this field when deserializing the collection. We could also just let the user decide what the type is which is easier, and I don't know how valuable inferring the type is anyway. That's why I am saying it doesn't really matter that it says `as T` since types in JS/TS are just for developer experience anyway

osalkanovic: I'm not sure that this solution here is 100% okay, but something in this direction could go: 

https://www.typescriptlang.org/play?#code/LAKFGMBsEMGdYAQGFoCcDKAHApuBBvUBYhcAewDtYAXVAV3GrNQApM6AjSASzwugC22AFwIaqbhQDmASgIBfUIrAqocRAFUKzACbZU2HQFlomADwAVAHwEiJclVoMmrdl14ID0HZUgBPBEkaaApwEQQLOUIQEgRlO2IpbGoWGVELBAAfBAo6SEhbGNiSSGSEAA8EAF4CfiFRAHJoBuVikgB6doABalgAWm4pbQME4oNqOlQKHOwAdwRqAAtuWAA6IOoQsJZymVHY0eV4kFLqBAFTapn5rV19QxNMFhQMHHAZAG4IShoEAJqLphVkkUnsQA5YGRSqtIGQpCwAhsttgyAAzZBoLC4AA0fwA-Ks6tgwUA

I don't know if it's okay to ask the user to pass us an instance of the class to constructor :D

volovyks: Allowing partial deserialization (simple `as DataType`) is also an option. Maybe JS developers are used to this behavior and do not expect functions to be in place.

petarvujovic98: @volovyks @osalkanovic Speaking as a JS dev, this behavior is totally fine. One convenience that could be added to something like this, because of the nature of serializing classes and collections could be something like the following:
```typescript
get(key: Bytes, deserializer?: (value: unknown) => T): T | null {
  const indexRaw = getIndexRaw(this.keyIndexPrefix, key);

  if (indexRaw) {
    const index = deserializeIndex(indexRaw);
    const value = this.values.get(index);
    
    if (value) {
      return !!deserializer ? deserializer(value) : (value as T);
    } else {
      throw new Error(ERR_INCOSISTENT_STATE);
    }
  }

  return null;
}
```

osalkanovic: @petarvujovic98  Nice, I agree. 
If @volovyks  and @ailisp  agree, I would correct and update the PR.

osalkanovic: @petarvujovic98 what are you think about this [example](https://github.com/near/near-sdk-js/pull/203#discussion_r964503394)?

petarvujovic98: @osalkanovic I wouldn't use this approach as the instance is going to also get serialized so you will lose the constructor, plus if it is a primitive type it won't need to have a constructor and you are complicating the DX. 

Keeping track of type data after seralization is difficult in JS, but since users are using TS to write the construction and access of variables after deserialization, it's better to keep the logic for those in the TS layer rather than in the JS/runtime layers.

ailisp: I agree with @petarvujovic98 , and the way @osalkanovic shown is quite clever that I didn't think of.  May @osalkanovic 's way be incorporated in as a pattern to write `deserializer?: (value: unknown) => T`

src/collections/unordered-map.ts
ailisp: :+1:

src/collections/unordered-map.ts
osalkanovic: @ailisp I will update PR with deserializer today.

src/collections/unordered-map.ts
```diff
@@ -59,21 +59,21 @@ export class UnorderedMap {
     return keysIsEmpty;
   }
 
-  get(key: Bytes): unknown | null {
+  get(key: Bytes, deserializer?: (value: unknown) => DataType): DataType | null {
```
osalkanovic, line 62: @petarvujovic98 since I see that you are working on the default value, maybe it would be good to convert this to options

src/collections/unordered-map.ts
osalkanovic: @ailisp @volovyks @petarvujovic98 PR is ready for review

src/collections/unordered-map.ts
petarvujovic98: @osalkanovic Seems reasonable to me. This also allows for more extensibility in the future

ailisp (approved):
Overall your work looks great! Thanks!

I'd like to add a small naming comment: in near-bindgen.ts, we call the process from object to DataType as "reconstruct", and deserializer in libraries usually mean from string/bytes to object. So I wonder if we should name it as `reconstructor` instead of `deserializer`? 

The other way is make deserializer calls on `raw` instead of on `value`, this allow user to even override the `JSON.parse` with a customizable format, but this seems to be more scoped than this PR intended and a `serializer?` argument is better to come together.

Therefore I suggest naming it as `reconstructor`, and make a room for future adding `deserializer` the get by changing the signatures from:
```ts
    get(key: Bytes, reconstructor?: (value: unknown) => DataType): DataType | null {
```
to
```ts
    get(key: Bytes, options?: {reconstructor?: (value: unknown) => DataType}): DataType | null {
```

And, could you please change a test in tests/src/unordered-map.ts to ensure get with option works. We have an existing test that does manually reconstruct: https://github.com/near/near-sdk-js/blob/ef90663432afa8d7d5b9ad54988359d8c931e7ad/tests/src/unordered-map.js#L61

With your PR a house will be get directly!

osalkanovic:
@ailisp Updated with latest develop, ready for merge! :)

ailisp:
@osalkanovic Fantastic! You have even implemented generics for the new lookup map. I think this PR is in production-ready quality. @volovyks If this also looks good to you, let's wait NEARCON end (at Sep 14 and merge this!)

ailisp:
@osalkanovic We turned on ci, if you could resolve and merge latest develop into PR. If everything is good on CI we can merge it!

osalkanovic:
@ailisp Ready for merge

ailisp:
@osalkanovic Thank you!

PR #204, opened 2022-09-06
`private` and  `payable` functionality

volovyks:
- `payableFunction` nad `privateFunction` parameters added to the @call decorator
- both `False` by default
- `privateFunction` is used instead of `private` because `private` is a reserved word
- for the sake of consistency and extendability all decorators accept parameters


ailisp:
The new api design on `call` and `view` looks great!

volovyks:
Thanks for the fix @ailisp!

ailisp (approved):
Awesome implementation!

PR #206, opened 2022-09-07
add includeBytes "macro" that helps factory contract

ailisp:
Address #189 

diff was too big to show in github.

before:
image

after:
image


volovyks (approved):
Cool! Should we include this updated babel.config in all the projects? Ot it is expected to be used rearly?

ailisp:
> Cool! Should we include this updated babel.config in all the projects? 

Yes it should in for all projects, is there a way to do it in near-sdk-js side?

volovyks:
Now we need to include it manually. But it's added automatically if the project is created with create-near-app.
@gagdiez let's not forget to add it.

PR #208, opened 2022-09-08
prommiseResult throws an error

volovyks:
We should not mimic the Rust design here. Throwing an error is the standard behavior for JS.
The same goes for the return type. I do not want to add {error, result} because it would be the only place in the whole API where we use such a design.

### DIFF
--- a/lib/api.d.ts
+++ b/lib/api.d.ts
@@ -1,5 +1,4 @@
 import { Bytes } from "./utils";
-import { PromiseResult } from "./types";
 export declare function log(...params: any[]): void;
 export declare function signerAccountId(): string;
 export declare function signerAccountPk(): Bytes;
@@ -50,7 +49,7 @@ export declare function promiseBatchActionDeleteKey(promiseIndex: number | bigin
 export declare function promiseBatchActionDeleteAccount(promiseIndex: number | bigint, beneficiaryId: string): void;
 export declare function promiseBatchActionFunctionCallWeight(promiseIndex: number | bigint, methodName: string, args: Bytes, amount: number | bigint, gas: number | bigint, weight: number | bigint): void;
 export declare function promiseResultsCount(): bigint;
-export declare function promiseResult(resultIdx: number | bigint): Bytes | PromiseResult.NotReady | PromiseResult.Failed;
+export declare function promiseResult(resultIdx: number | bigint): Bytes;
 export declare function promiseReturn(promiseIdx: number | bigint): void;
 export declare function storageWrite(key: Bytes, value: Bytes): boolean;
 export declare function storageRemove(key: Bytes): boolean;
--- a/lib/api.js
+++ b/lib/api.js
@@ -140,7 +140,6 @@ export function accountLockedBalance() {
     return env.account_locked_balance();
 }
 export function valueReturn(value) {
-    log('valueReturn');
     env.value_return(value);
 }
 export function promiseCreate(accountId, methodName, args, amount, gas) {
@@ -196,16 +195,12 @@ export function promiseResult(resultIdx) {
     if (status == PromiseResult.Successful) {
         return env.read_register(0);
     }
-    else if (status == PromiseResult.Failed ||
-        status == PromiseResult.NotReady) {
-        return status;
-    }
     else {
-        throw Error(`Unexpected return code: ${status}`);
+        throw Error(`Promise result ${status == PromiseResult.Failed ? "Failed" :
+            status == PromiseResult.NotReady ? "NotReady" : status}`);
     }
 }
 export function promiseReturn(promiseIdx) {
-    log('promiseReturn');
     env.promise_return(promiseIdx);
 }
 export function storageWrite(key, value) {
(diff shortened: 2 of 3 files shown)


ailisp (approved):
Nice! Didn't think of that, the throwing-Error or return the success value just beautifully fit!

gagdiez:
@ailisp @volovyks I would still argue in favour of having a {status, result}, as JS promises do, since they are simpler to explain to developers. 

Lets compare a simple workflow, where the callback returns the result value:

#### result structure
```js
let response = near.promiseResult();

if(response.status === success){
  log("it worked")
  return response.value
}else{
  log("it didn't work because:", response.error)
  return
}
```

#### try/catch
```js

try{
  let result = near.promiseResult(); // if this fails the catch arises
  log("it worked")
  return result
}catch(e){
  log("it didn't work because:", e.message)
  return
}
```

I personally find the first logic simpler to read and explain for devs.

gagdiez:
@ailisp @volovyks we could even use the status `pending`, `fulfilled` and `rejected` for the promise result, as JS promises, and the result can be called `value`

volovyks:
I think that the concept of `{status, value}` is more confusing. The developer will need to check the function signature, or literally read the source code if he is using JS.

Let's not forget that `try{} catch() {}` is optional and you can just
```JS
return near.promiseResult() // bad practice, use try-catch!
```

Looks like we need  1-2 more opinions in order to make a decision. @BenKurrek @doriancrutcher @petarvujovic98

ailisp:
I vote this PR's throw behavior. Even if the developer forget to write try-catch, the throw will cause contract panic and there's not going to be a security issue. Also, throwing an error is more common practice in JavaScript libraries than return `{error, value}`

volovyks:
Ok, merging. We can always come back to this question if developers start complaining.

gagdiez:
@ailisp @volovyks I was thinking yesterday, and there is another thing I don't like about the try/catch pattern: 
it will catch **any error**, and not only the callback failure.

```js
try{
  let result = near.promiseResult(); // if this fails the catch arises
  // but if I have a bug/error here also!
}
```

If I introduced a bug in my callback method (for example, accessing the wrong index), it will be masked by the try/catch.

This can be even dangerous in the classic situation where: 
1. the user calls a method in my contract attaching money, I make state changes
2. I make a XCC sending money
3. I check if it failed in the callback, and rollback state + send money back if necessary

If my callback has an error, the code will act as if the xcc failed and send money back to the user.

The solution to that is using:

```ts
let result: type = undefined;
let xcc_succeeded: bool = false;
try{
  let result = near.promiseResult(); // if this fails the catch arises
  xcc_succeeded = true;
} catch(e){ }

if (succeeded){
  // it succeeded: do something
}else{
  // it failed: rollback
}
```

since in that way the only way the rollback executes is if the xcc failed.

Furthermore, internal errors are not masked, if I made a mistake, the contract will panic with the correct error message.

Because of this, I think the `try/catch` patter can be error-prone, and it would be simpler and safer to use a simple structure with a `{status, result}`.

osalkanovic:
Maybe having both options can cover all cases
Example is [here](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgAoCUIGcCuAbMZAbwFgAoZS5KbfMALmRBwFsAjaAGnKuQDc4eHBEZYwUUAHMA3OQC+5chGYs0UAPYtgWCJlwFiPKgDl1YTHAAmAT2QBeZAAZuFKgGUcCJFiwx895ABGF14AMThgPAhLAIAmFwUyRTI-EAQwYHUQZAAHDS0dPToAChp9MABJSwAPRhUOKGQAH2Q2YElQME5keEi3SOUwa0ZxYQBKRgxaAlkUnDSMrNz87V1psFL1qtqmVgbm1vbO7t68fqjwawB+Rng8HQnkMQkQGWSe+fTM7LzNVaKCJtyts6ntoAc2h1wCcImcBpcbq11OoonAQGNDK4qFFCGI4GAcFhJitCusAoEjJRgDBiniCVh7HYHKgSWtygA6DxebC+fBjUhY3jIBBZMT8QTCUTiKQBABEuG5PlllKFNAJUGyp3Og1sV0oRAEQhE4qN3TKdEYgTkyEYhuEVBVcgg9xQAqFVGpxS18KG-JV7soapwGuIduNstOsrN60YjkSAeQTpdmITvDAAAsNAB3JgQHMAUSgGigxQA5OyK+zS2N-VR43X3pRKUZ5MkRSAxZAxQ5fgU2SVnMhRhAa2R253sGBAgFe-91sVHKPyOOURB2Xh1JJil2ukPJ4ExkA)

petarvujovic98:
I think that the `{ status, result }` pattern is pretty common in JS. For reference the standard `fetch` method has this for HTTP requests, so developers are pretty familiar with this. 

I also think that the `throw` does feel more JS-like, but not all JS features are good, even if they are common practice. So I would vote for the `{ status, result }` pattern here like @gagdiez.

Issue #191, opened 2022-08-23
Error Message Request: Existing info in LookupMap


doriancrutcher:
Let's say you have a lookup map in your smart contract and you set the key as `name` and the value as `Dorian` 

Then you run the method again placing the same value `Dorian` under the key `name` 

the current error message received is 

```
"Smart contract panicked: unexpected token: 'undefined'\n"
```

Which is a bit misleading to what's actually going on. 

Could the message be something like `value already exists`?

ailisp:
This looks like a bug, we're expecting `set` to work when value already exists

PR #209, opened 2022-09-08
LookupMap set twise test added

volovyks:
I was not able to reproduce the error described in https://github.com/near/near-sdk-js/issues/191, looks like it works. @doriancrutcher we can still merge this PR, but please, tell me if you are facing this problem in some edge cases.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -140,7 +140,6 @@ export function accountLockedBalance() {
     return env.account_locked_balance();
 }
 export function valueReturn(value) {
-    log('valueReturn');
     env.value_return(value);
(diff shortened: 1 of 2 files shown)


ailisp (approved):
Looks this is a solid test. @doriancrutcher could you share your contract that was error in #191 ?

ailisp (approved):
Looks like this is a solid test. @doriancrutcher could you share your contract that was error in #191 ?

volovyks:
@doriancrutcher merging this now, feel free to reopen if there are still some issues with this.

PR #211, opened 2022-09-08
update readme to include recently added features. correct some api types

ailisp:
Mostly, I added high level promise apis, includeBytes and PublicKey types. Change `BigInt` to `bigint` and correct a few typing in api.ts

@volovyks Could you write a section on your awesome work on `@nearbindgen`, `@call`, `@view` and also describe the init / default behavior? I think it can be between `Alt BN128` and  `collections`, but if you have a better idea on organize the do it will be cool. Thanks!


volovyks:
@ailisp, I have added the docs for NearBingden and other decorators.

volovyks (approved):

Issue #107, opened 2022-06-18
Remove need for redundant reads/writes for `UnorderedMap`


austinabell:
Seems like the implementation was taken from the old Rust SDK algorithm. The inefficiency is that it stores a mapping of key to index, one vector for keys and one for vector.

So for every read/write, you are charging basically twice for each read/write since they are separated instead of just being one vector of key/value tuples (`<span class="error">[string, string]</span>`). This is an easy optimization but is state-breaking.

The second inefficiency is having to do two storage reads when retrieving any value because the indices are stored in a separate lookup. This is also inefficient because you do swap_removes when removing entries, and this means that two different storage reads need to be charged for this operation.

The optimized version (which may not be perfect, but what we decided to go with) is here [https://github.com/near/near-sdk-rs/blob/1a39f1e2406517ee93bc4efb0b859751297e645d/near-sdk/src/store/unordered_map/mod.rs](https://github.com/near/near-sdk-rs/blob/1a39f1e2406517ee93bc4efb0b859751297e645d/near-sdk/src/store/unordered_map/mod.rs). The `FreeList` type that is used isn't required, but removes the requirement of having to swap_remove values.

My recommendation would be to optimize this sooner than later before it starts being used in prod, or else you will be stuck with this inefficiency to avoid making state-breaking changes (like we are with the collection this algo was taken from).

ailisp:
Got it! You're right, this is a high priority issue that we should address now

ryancwalsh:
This does sound important. Does anyone have plans to address it in the next couple of days as NEARCON approaches? Or is it lower priority?

ailisp:
I'll have a look today!

petarvujovic98:
Here is a [repository](https://github.com/Learn-NEAR/NCD.L1.sample--lottery) I was working on, that is experiencing some gas issues. This is a semi-intermediate example yet the collection inefficiencies are causing gas exceeds in my tests.

At first I thought it might be Promises that were causing this but upon further debugging it seems that the `UnorderedSet` was the culprit. Although there might also be other problems.

This is also a contract which was ported from AssemblyScript to both Rust and JavaScript and only JavaScript is having these issues with gas. I think it is pretty important to address this as it means that in any complex contract the SDK might be unusable

ailisp:
@petarvujovic98 I debugged your contract, looks like it's a different issue: the promise object need to return instead of call onReturn, and tests passed after this change:

[...]

PR #212, opened 2022-09-09
(state breaking) make get set in unorderedmap more efficient

ailisp:
This is a state breaking change, addresses #107, inspired by @austinabell 's suggestions

I assume developers contract using unorderedmap is still in testnet / no production users and safe to clear the state to use the state breaking change. If this is not the case, then you can use the following strategy to upgrade state:
1. write a one-method contract with old near-sdk-js, iterate and get all key-value pairs in your unordered map
2. clear states that are prefixed by the old unordered map prefix
3. write a one-method temp contract with new near-sdk-js, add a method to add all (key, value) to the new unordered map
4. remove the temp contract, build your contract with new-sdk-js and deploy!

### DIFF
--- a/lib/collections/unordered-map.d.ts
+++ b/lib/collections/unordered-map.d.ts
@@ -1,13 +1,12 @@
 import { Bytes } from "../utils";
 import { Vector } from "./vector";
+import { LookupMap } from "./lookup-map";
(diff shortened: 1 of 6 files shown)


volovyks (approved):
Nice simplification!

PR #218, opened 2022-09-14
rename plain object -> class instance operation to "reconstruct"

ailisp:
I think this naming describes the signature better: from plain object to class instance. Typically deserialize mean from bytes to object. Also aligns to #203 .


volovyks (approved):
Good point.
Let's merge it tomorrow.

PR #219, opened 2022-09-14
check in ci that yarn build generated files are included/updated

ailisp:
We didn't have a check to ensure changes on ts files are built and committed. So I add this check

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,19 @@
+name: Build
+on: push
+jobs:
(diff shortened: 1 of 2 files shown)


volovyks (approved):
Great! So, it will fail if after `yarn build` we will have something in `git diff` right?

ailisp:
@volovyks Yes. git status must be clean

volovyks:
@ailisp is the change in yarn.lock intended?

PR #220, opened 2022-09-14
always kill sandbox process

volovyks:
Before this change, we were not stopping near-sandbox processes in case the test failed. With time it leads to high memory, CPU, and battery usage.
afterEach.always(...) solves the problem.


ailisp (approved):
Looks good!

PR #221, opened 2022-09-15
run ci also on pull request for external contributors

ailisp:


volovyks (approved):
Wow, not an obvious logic!

PR #225, opened 2022-09-15
Windows tests turned on (TEST)

volovyks:


ailisp:
Great attempt. I didn't know github actions support windows! The error shows we need to build dependant binary tools for windows first?

volovyks:
@ailisp it was an error thrown by us. I have improved our post-install script and whitelisted windows. Now it is throwing the right error.

Issue #172, opened 2022-08-10
BigInt serialization support


austinabell:
The `BigInt` type used throughout the library errors on `JSON.serialize` ([by design](https://github.com/GoogleChromeLabs/jsbi/issues/30)) but is a footgun for developers.

My intuition is that something like [the suggestion on the linked issue](https://github.com/GoogleChromeLabs/jsbi/issues/30#issuecomment-521449285), except just including the receiver function to the JSON.stringify, might be sufficient for our internal JSON serializing of state.

The issue is that deserializing might not be so straightforward since, from a string we may not be able to infer the return type based on the `JSON.parse` function signature.

JS is so wonderful. Maybe if there is a way at the SDK level to override the stringify and parse methods of BigInt, similar to what [this comment](https://github.com/GoogleChromeLabs/jsbi/issues/30#issuecomment-1006088574) suggests, might be the cleanest?

ailisp:
I was just about to create the same issue. Thanks!

> JS is so wonderful. Maybe if there is a way at the SDK level to override the stringify and parse methods of BigInt, similar to what https://github.com/GoogleChromeLabs/jsbi/issues/30#issuecomment-1006088574 suggests, might be the cleanest?

Good idea!

PR #213, opened 2022-09-09
Refactor the project and add generics and getter options

petarvujovic98:
I used basic eslint and prettier installs without any configs to format the files.

There were also function refactors for better maintainability and JS standards as well as variable renames.

I also added generics to collections as well as a optional getter options object which has a reconstructor and default value property.

PR #228, opened 2022-09-21
feat: Add BigInt serialization

petarvujovic98:
Add custom serialization function in utils
Improve types in near-bindgen and use serialization function Add separate serialization for function returns
Add serialization to collections
Add serialization tests
Add serialization build and tests scripts

Addresses #172.

This is just a design I came up with that we can iterate on.

### DIFF
--- a/src/collections/lookup-map.ts
+++ b/src/collections/lookup-map.ts
@@ -1,6 +1,6 @@
 import * as near from "../api";
 import { GetOptions } from "../types/collections";
-import { Bytes, getValueWithOptions } from "../utils";
+import { Bytes, deserialize, getValueWithOptions, serialize } from "../utils";
 
 export class LookupMap<DataType> {
   constructor(readonly keyPrefix: Bytes) {}
@@ -12,7 +12,7 @@ export class LookupMap<DataType> {
 
   get(key: Bytes, options?: GetOptions<DataType>): DataType | null {
     const storageKey = this.keyPrefix + key;
-    const value = JSON.parse(near.storageRead(storageKey));
+    const value = deserialize(near.storageRead(storageKey));
 
     return getValueWithOptions(value, options);
   }
@@ -24,7 +24,7 @@ export class LookupMap<DataType> {
       return options?.defaultValue ?? null;
     }
 
-    const value = JSON.parse(near.storageGetEvicted());
+    const value = deserialize(near.storageGetEvicted());
 
     return getValueWithOptions(value, options);
   }
@@ -35,13 +35,13 @@ export class LookupMap<DataType> {
     options?: GetOptions<DataType>
   ): DataType | null {
     const storageKey = this.keyPrefix + key;
-    const storageValue = JSON.stringify(newValue);
+    const storageValue = serialize(newValue);
 
     if (!near.storageWrite(storageKey, storageValue)) {
       return options?.defaultValue ?? null;
     }
 
-    const value = JSON.parse(near.storageGetEvicted());
+    const value = deserialize(near.storageGetEvicted());
 
     return getValueWithOptions(value, options);
   }
@@ -56,7 +56,7 @@ export class LookupMap<DataType> {
   }
 
   serialize(): string {
-    return JSON.stringify(this);
+    return serialize(this);
   }
 
   // converting plain object to class object
--- a/src/collections/lookup-map.ts
+++ b/src/collections/lookup-map.ts
@@ -1,6 +1,10 @@
 import * as near from "../api";
 import { GetOptions } from "../types/collections";
-import { Bytes, deserialize, getValueWithOptions, serialize } from "../utils";
+import {
+  Bytes,
+  getValueWithOptions,
+  serializeValueWithOptions,
+} from "../utils";
 
 export class LookupMap<DataType> {
   constructor(readonly keyPrefix: Bytes) {}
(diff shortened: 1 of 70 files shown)


src/collections/lookup-map.ts
```diff
@@ -12,7 +12,7 @@ export class LookupMap<DataType> {
 
   get(key: Bytes, options?: GetOptions<DataType>): DataType | null {
     const storageKey = this.keyPrefix + key;
-    const value = JSON.parse(near.storageRead(storageKey));
+    const value = deserialize(near.storageRead(storageKey));
```
ailisp, line 15: It's great to make serialize / desalinize general / independent of the collections! Since we have `GetOptions`, it's possibly a good idea to make it as part of `GetOptions`, and have current `utils.serialize`/`utils.deserialize` as the default option.

src/utils.ts
```diff
@@ -63,3 +66,32 @@ export function getValueWithOptions<DataType>(
 
   return value as DataType;
 }
+
+export function serialize(valueToSerialize: unknown): string {
+  return JSON.stringify(valueToSerialize, (_, value) => {
+    if (typeof value === "bigint") {
+      return {
+        value: value.toString(),
+        [BIGINT_KEY]: BIGINT_BRAND,
```
ailisp, line 75: This is very clever mechanism! A little extension based on your approach can become more general. We don't need both BIGINT_KEY and BIGINT_BRAND to indicate it's a bigint, so, maybe it can be TYPE_KEY and TYPE?

ailisp (approved):
I like your solution, it solves the bigint serialization in a clean, idiomatic way. I leave some comments to extend your framework a little bit. /cc @volovyks

src/collections/lookup-map.ts
petarvujovic98: @ailisp I have covered this in the latest commits. Let's discuss the approach I took there. I also added a `serializer` and `deserializer` option to the `NearBindgen` decorator

src/utils.ts
petarvujovic98: @ailisp Could you perhaps give an example or code snippet of what you are thinking about when you say _more general_?

src/collections/lookup-map.ts
ailisp: That looks great!

petarvujovic98:
@volovyks @ailisp We should change examples and tests to use the new serialization, and to use the options we added in #213.
Do we want to do this in this PR or create a separate PR?

src/utils.ts
ailisp: For example, if later we also want to support serialize and deserialize Date, we can extend your serialize code for date:
```js
return {
  value: value.toString(),
  [TYPE_KEY]: DATE_TYPE,
}
```

And in deserialization, we can always check `TYPE_KEY`, see what type it is and call bigint / date deserialization correspondingly

petarvujovic98: @ailisp I added something similar along with an implementation of `Date` serialization and an accompanying test in [this commit](https://github.com/near/near-sdk-js/pull/228/commits/d843af664b8b1bdc70719b859a3715cf63d4a884)

ailisp (approved):
Wow you're awesome! I was just mention Date as an imaginary example and you've already made it with ! It will be helpful too, developers will be glad to serialize and deserialize date automatically. 

@volovyks If you also agree with the change I think we can merge it

src/utils.ts
```diff
@@ -50,16 +60,71 @@ export function assert(expression: boolean, message: string): void {
...
+  // @ts-ignore
+  Date.prototype.toJSON = function () {
+    return {
+      value: this.toISOString(),
+      [TYPE_KEY]: TypeBrand.DATE,
+    };
+  };
+
+  return JSON.stringify(valueToSerialize, (_, value) => {
+    if (typeof value === "bigint") {
```
ailisp, line 101: I think it's better to check type isinstance of Date instead of override `Date.prototype`?

tests/__tests__/test-date-serialization.ava.js
```diff
@@ -0,0 +1,51 @@
...
+  const dateField = await dsContract.view("getDateField");
+  t.is(dateField, new Date(0).toISOString());
+
+  const newDate = new Date();
+  await dsContract.call(dsContract, "setDateField", { dateField: newDate });
+  const afterSet = await dsContract.view("getDateField");
+  t.is(afterSet, newDate.toISOString());
+});
+
+test("get date field in milliseconds", async (t) => {
```
ailisp, line 42: These tests are great :+1:

src/utils.ts
petarvujovic98: Unfortunately, by the time you get the value argument inside the `stringify`replacer, the `toJSON` function of the `Date`object is already called so I cannot infer the type of the variable, I tried several ways to make it happen but this was the only one that worked, evem though I don't like it's hacky nature/playing around with prototypes

ailisp: I see, thanks for clarification!

ailisp: Also doesn't have luck with value - as you said it's already string, but I found this works: https://stackoverflow.com/a/54037861

petarvujovic98: Thank you. I just updated the `Date` serialization function with this suggestion and it seems to work as expected.

volovyks (approved):
It was a joy to review it, thanks @petarvujovic98 !

PR #249, opened 2022-09-29
Fix #247

### LINKED ISSUE #247: Dust in attachedDeposit
with attachedDeposit
exam: 5 Near
When I make a function call that gets the amount I attached, I always get a number greater than 5

Why does some dust appear?
2594073385365405696
but if i attach 10 near, this dust number does not appear

I think this is a strange problem
It happens with 5 but not with 10
I tried some other numbers and some numbers come up and some don't
unknown
unknown1
unknown2

ailisp:
Fix #247. Main fix is here: https://github.com/near/quickjs/pull/4. Extend the test to verify the fix works.

### DIFF
--- a/cli/post-install.js
+++ b/cli/post-install.js
@@ -37,7 +37,7 @@ await executeCommand(`rm ${BINARYEN_TAR_NAME}`);
 
 console.log("Installing QuickJS...");
 
-const QUICK_JS_VERSION = `0.1.1`;
+const QUICK_JS_VERSION = `0.1.2`;
 const QUICK_JS_VERSION_TAG = `v${QUICK_JS_VERSION}`;
 const QUICK_JS_SYSTEM_NAME =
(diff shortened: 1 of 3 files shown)


volovyks (approved):
Great investigation on this issue!
You mentioned that the result could depend on the number itself. Should we test not just zero, but several more numbers just to be sure?

ailisp:
> You mentioned that the result could depend on the number itself. Should we test not just zero, but several more numbers just to be sure?

I tested [attached deposit 1-10 NEAR](https://github.com/near/near-sdk-js/pull/249/files#diff-e4a3ec83b51bb6eccb543bc44266208daf4baf1c781216a759405d8e24ddbcfdR108). Previously, some is correct, some is not. After the fix, all of them work, which probably can mean it fixes the wrong cases while still correct in the correct cases.

PR #251, opened 2022-09-30
Add nested and nested nested collection example

ailisp:
@volovyks @petarvujovic98 @osalkanovic This is the way that I can figure out with recent enhancement in collection APIs. It's relatively intuitive, one way that work for all collections and support any number of nested collections. Of course there's more than one way of doing this now and it's possible that you come up with a better way. Please feel free to suggest or make another PR to discuss!


volovyks (approved):
@ailisp I have tried to come up with a better version, but looks like it is not possible with the current collections APIs. We can come back to it someday after 1.0.

PR #253, opened 2022-10-04
Add Storage Management to FT Example

idea404:
This PR adds storage management to the `fungible-token` example. 

It also changes the naming of public functions to `snake_case` thus matching naming conventions for Rust smart contracts.


examples/src/fungible-token.js
```diff
@@ -1,87 +1,136 @@
...
 @NearBindgen({ initRequired: true })
 export class FungibleToken {
-  constructor() {
-    this.accounts = new LookupMap("a");
-    this.totalSupply = 0n;
-  }
+  accounts = new LookupMap("a");
+  accountRegistrants = new LookupMap("r");
+  accountDeposits = new LookupMap("c");
+  totalSupply = "0";
```
volovyks, line 8: Any reason to change it to string?

examples/src/fungible-token.js
```diff
@@ -1,87 +1,136 @@
...
 
   @initialize({})
-  init({ prefix, totalSupply }) {
-    this.accounts = new LookupMap(prefix);
-    this.totalSupply = BigInt(totalSupply);
-    this.accounts.set(near.signerAccountId(), this.totalSupply);
-    // In a real world Fungible Token contract, storage management is required to denfense drain-storage attack
+  init({ owner_id, total_supply }) {
+    assert(BigInt(total_supply) > BigInt(0), "Total supply should be a positive number");
+    assert(this.totalSupply === "0", "Contract is already initialized");
```
volovyks, line 13: Functions flagged with @initialize can be executed only once. I guess this check will never be true.

examples/src/fungible-token.js
```diff
@@ -1,87 +1,136 @@
...
+  totalSupply = "0";
 
   @initialize({})
-  init({ prefix, totalSupply }) {
-    this.accounts = new LookupMap(prefix);
-    this.totalSupply = BigInt(totalSupply);
-    this.accounts.set(near.signerAccountId(), this.totalSupply);
-    // In a real world Fungible Token contract, storage management is required to denfense drain-storage attack
+  init({ owner_id, total_supply }) {
+    assert(BigInt(total_supply) > BigInt(0), "Total supply should be a positive number");
```
volovyks, line 12: This a good move! Let's check owner_id as well (we have a function for that in utils).

examples/src/fungible-token.js
```diff
@@ -1,87 +1,136 @@
...
+    assert(this.totalSupply === "0", "Contract is already initialized");
+    this.totalSupply = total_supply;
+    this.accounts.set(owner_id, this.totalSupply);
+  }
+
+  internalGetMaxAccountStorageUsage() {
```
volovyks, line 18: Do we really need to actually add and delete acc to get max storage?
Are we mimicking rust logic here, is it done there the same way?
If we really need to get storage usage, why not pass accId and require the user to attach less N for shorter accs?

examples/src/fungible-token.js
```diff
@@ -1,87 +1,136 @@
...
-    let balance = this.accounts.get(accountId, { defaultValue: 0n });
-    let newBalance = balance - BigInt(amount);
-    assert(newBalance >= 0n, "The account doesn't have enough balance");
-    this.accounts.set(accountId, newBalance);
-    let newSupply = this.totalSupply - BigInt(amount);
-    assert(newSupply >= 0n, "Total supply overflow");
-    this.totalSupply = newSupply;
+  internalSendNEAR(receivingAccountId, amountBigInt) {
+    assert(amountBigInt > BigInt("0"), "The amount should be a positive number");
+    assert(receivingAccountId != near.currentAccountId(), "Can't transfer to the contract itself");
```
volovyks, line 36: Why not? IS it a part of the standard?

examples/src/fungible-token.js
```diff
@@ -1,87 +1,136 @@
...
+    assert(amountBigInt < near.accountBalance(), `Not enough balance ${near.accountBalance()} to cover transfer of ${amountBigInt} yoctoNEAR`);
+    const transferPromiseId = near.promiseBatchCreate(receivingAccountId);
```
volovyks, line 38: Try to use high-level promise API here. We have examples in the repo.

ailisp (approved):
Thanks @idea404 . This is a great addition to FT to avoid exhaust-storage-deposit attack. Can you fix the lint so we can merge it?

examples/src/fungible-token.js
```diff
@@ -1,87 +1,136 @@
...
-  internalWithdraw({ accountId, amount }) {
-    let balance = this.accounts.get(accountId, { defaultValue: 0n });
-    let newBalance = balance - BigInt(amount);
-    assert(newBalance >= 0n, "The account doesn't have enough balance");
-    this.accounts.set(accountId, newBalance);
-    let newSupply = this.totalSupply - BigInt(amount);
-    assert(newSupply >= 0n, "Total supply overflow");
-    this.totalSupply = newSupply;
+  internalSendNEAR(receivingAccountId, amountBigInt) {
+    assert(amountBigInt > BigInt("0"), "The amount should be a positive number");
```
volovyks, line 35: Let's move asserts to a function if we are using it more than once.

examples/src/fungible-token.js
```diff
@@ -1,87 +1,136 @@
...
-    assert(amountInt > 0n, "The amount should be a positive number");
-    this.internalWithdraw({ accountId: senderId, amount });
-    this.internalDeposit({ accountId: receiverId, amount });
+    assert(BigInt(amount) > BigInt(0), "The amount should be a positive number");
+    this.internalWithdraw(senderId, amount);
+    this.internalDeposit(receiverId, amount);
+  }
+
+  @call({ payableFunction: true })
+  storage_deposit({ account_id }) {
```
volovyks, line 74: Hm... why are we using JS for this example? Types will be beneficial here.

volovyks (commented):
@idea404 thanks for working on this!
nits:
- please, remove package.lock file, yarn.lock is enouth
- changes to yarn.lock are not necessary, since there is no changes in package.json
- linter is not happy

examples/src/fungible-token.js
idea404: I was getting BigInt -> string errors in the `_serialize()` function within `/build` during deployment in tests. A line similar to this one: 

```js
  return JSON.stringify(value);
```

examples/src/fungible-token.js
idea404: Good point, removed

examples/src/fungible-token.js
idea404: Added! 🙌

examples/src/fungible-token.js
idea404: > Do we really need to actually add and delete acc to get max storage?

There is a way by getting the cost per character and then multiplying by max length of accountId (64 chars), but this implementation is also seen here: 

https://github.com/near/near-sdk-rs/blob/1a39f1e2406517ee93bc4efb0b859751297e645d/near-contract-standards/src/fungible_token/core_impl.rs#L50-L56

examples/src/fungible-token.js
idea404: Can't think of any strong reason, was aiming to isolate the use of this function specifically to what I intended when writing that assert. Will remove it now as I see no harm in not having that there.

examples/src/fungible-token.js
idea404: Good point actually. Will adapt to TS and request re-review

examples/src/fungible-token.js
idea404: sounds good, have abstracted as static methods in an `Assertions` class

examples/src/fungible-token.js
idea404: Sounds good: implemented

examples/src/fungible-token.js
idea404: Implemented dynamic storage on accId length

examples/src/fungible-token.js
```diff
@@ -1,87 +1,136 @@
...
-  initialize,
-  near,
-  LookupMap,
-  assert,
-} from "near-sdk-js";
+import { NearBindgen, call, view, initialize, near, LookupMap, assert } from "near-sdk-js";
 
 @NearBindgen({ initRequired: true })
 export class FungibleToken {
-  constructor() {
```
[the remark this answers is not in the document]
volovyks: Constructor is required for all contracts that want to deserialize their members.
Check this: https://github.com/near/near-sdk-js/blob/develop/src/near-bindgen.ts#L151

examples/src/fungible-token.js
volovyks: BigInt serialization was added in this PR: https://github.com/near/near-sdk-js/pull/228
So it should work. Please, check it. If it will fail anyway - let's try to discuss and fix it here.

examples/src/fungible-token.js
volovyks: Cool!
I will clarify, why storage logic is so complex.

examples/src/fungible-token.js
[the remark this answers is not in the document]
idea404: implemented constructor

examples/src/fungible-token.js
idea404: maps save bigints now to state

PR #255, opened 2022-10-05
fix: node 16.6.0 issue

### LINKED ISSUE #15: Rollup fails witn Node v16.6.0
We should understand why this is happening and fix the bug or set supported versions of Node. Works fine with Node v16.14.2

Olyno:
## Description

This pull request should avoid Rollup to fail with Node 16.6.0. You can have details about this issue here: https://github.com/babel/babel/discussions/15015#discussioncomment-3801654

## Related issue

This closes #15.


ailisp:
Thanks! Do you have test it with some versions after and before 16.6.0? I suspect not only just 16.6.0 not work. Let's try testing a few of them on CI

ailisp (approved):
Approve to run ci

Olyno:
I tried a few more versions, and I didn't find any anomalies. It will be a bit complicated to test each version one by one. Could I recommend adjusting the versions according to the users' feedback?

ailisp:
@Olyno That's reasonable! Have you tested 16.5.0, 16.7.0, latest 17.x and latest 18.x?

Olyno:
I just tried the versions you asked for, and they all seem to work fine.

Should we add a security layer by adding a build worflow for these versions? By using Github actions, we could reduce the cases of incompatibility @ailisp

volovyks (approved):
Nice!
@Olyno I think that is a good idea. Let's add GitHub actions matrix with all major supported Node versions. But please, let's make it a separate job and run only 1 build and 1 test run because it will take forever to be completed.

PR #257, opened 2022-10-11
Add Babel config programmatically

petarvujovic98:
Remove the need to have a separate Babel config file with our custom plugins and presets and encode them programmatically.
Fix some ESLint errors from previous PRs.

This is mentioned in #229.

Note: This Babel config change will still enable the users of the SDK to add their own Babel config file if they need to, those config files will be merged with the config we pass while our config will have precedence.


volovyks (approved):
Awesome @petarvujovic98 !

PR #261, opened 2022-10-12
feat: Improve CLI output

petarvujovic98:
Transform CLI tools to TypeScript
Add CLI tsconfig file and update build script
Fix ESLint types resolution
Change argument parser from yargs to commander
Add signale logger
Add verbose flag
Add verbose flag recognition to near-bindgen-exporter

Addresses #227


petarvujovic98:
@volovyks @ailisp Unit tests failed because of hooks failing. I don't know what to do about this as sometimes it happens on my laptop as well, it isn't really predictable either. So this might be something we want to look into.

tsconfig.cli.json
```diff
@@ -0,0 +1,24 @@
+{
```
volovyks, line 1: Why do we need 2 tsconfig files?

cli/post-install.js
```diff
@@ -1,98 +1,64 @@
 import { executeCommand } from "./utils.js";
```
volovyks, line 1: Since now it's a build artifact - we can hide it from diff. Maybe even move it to `/dist` or `/lib`.

cli/package.json
```diff
@@ -0,0 +1,74 @@
+{
```
volovyks, line 1: Looks like copy pasted file

cli/src/cli/cli.js
```diff
@@ -0,0 +1,188 @@
...
+    await createHeaderFileWithQjsc(ROLLUP_TARGET, QJSC_TARGET, verbose);
+    signale.await("Generating methods.h file...");
+    await createMethodsHeaderFile(ROLLUP_TARGET, verbose);
+    signale.await(`Creating ${CONTRACT_TARGET} contract...`);
+    await createWasmContract(QJSC_TARGET, CONTRACT_TARGET, verbose);
+    signale.await("Executing wasi-stub...");
+    await wasiStubContract(CONTRACT_TARGET, verbose);
+    signale.success("Generated contract successfully!");
+}
+// yargs(hideBin(process.argv))
```
volovyks, line 67: Commented code :)

volovyks (commented):
@petarvujovic98  Looks good overall!

cli/post-install.js
petarvujovic98: @volovyks This ties in with [this comment](https://github.com/near/near-sdk-js/pull/261#discussion_r993457093).

cli/package.json
petarvujovic98: @volovyks You are right, didn't see that this happened.

cli/src/cli/cli.js
petarvujovic98: @volovyks Will remove!

tsconfig.cli.json
volovyks: Let's experiment with it in a separate PR. Maybe it's a good idea to move /cli to /src. Build artifacts should live separately.

ailisp (approved):
Great enhancement to DevX! Thank you @petarvujovic98

Issue #256, opened 2022-10-09
Middlewares


osalkanovic:
It would be good to have `@middleware` decorator

Example: 
```typescript
import {middleware, Middleware, near} from "near-sdk-js"

class OwnerValidator implements Middleware {
   
   call(data, next) {
     if(near.predecessorAccountId() != "test.near.org") //throw error
     next()
   }
}

....

@middleware(OwnerValidator)
@call()
create_user(){}

```

It has many advantages such as data validation, deposit validation, etc
I don't think it will be a big challenge to implement this.  If you guys agree, I'd love to implement this

@ailisp @volovyks

ailisp:
Looks like a nice addition to SDK! @volovyks what do you think?

volovyks:
Hm... I like the idea to automate common actions. But maybe some of them (like the one above), can be a parameter in @call decorator.
@osalkanovic do you have a link with the implementation? Can it live outside of the repo? (we can use it in our examples).

gagdiez:
@osalkanovic  is there something similar in the rust sdk?

osalkanovic:
> Hm... I like the idea to automate common actions. But maybe some of them (like the one above), can be a parameter in @call decorator. @osalkanovic do you have a link with the implementation? Can it live outside of the repo? (we can use it in our examples).

I dont have example, I just recommend as idea

osalkanovic:
> @osalkanovic is there something similar in the rust sdk?

I think no :D

petarvujovic98:
I think #263 closes this.

volovyks:
Yes, thank you @petarvujovic98
And thank you for the idea @osalkanovic

PR #263, opened 2022-10-13
[RFC] feat: Middlewares addition

petarvujovic98:
Add middlewares decorator
Add middlewares test

Keep in mind that right now the `middleware` decorator needs to be applied after the other function decorators to preserve the capturing of the respective methods. This, however, could be changed if we want to change the way we capture the decorated methods, e.g. by adding a _brand_ property to the decorated functions rather than looking for the name of the function decorator.

Addresses #256

### DIFF
--- a/src/near-bindgen.ts
+++ b/src/near-bindgen.ts
@@ -99,7 +99,7 @@ interface Middleware<Arguments extends Array<any>> {
    *
    * @param args - Arguments that will be passed to the function - immutable.
    */
-  call(...args: Arguments): void;
+  (...args: Arguments): void;
 }
 
 /**
@@ -123,7 +123,7 @@ export function middleware<Arguments extends Array<any>>(
     // @ts-ignore
     descriptor.value = function (...args: Arguments): ReturnType<AnyFunction> {
       try {
-        middlewares.forEach((middleware) => middleware.call(...args));
+        middlewares.forEach((middleware) => middleware(...args));
       } catch (error) {
         throw new Error(error);
       }
--- a/src/near-bindgen.ts
+++ b/src/near-bindgen.ts
@@ -90,6 +90,50 @@ export function call({
 }
 
 /**
+ * The interface that a middleware has to implement in order to be used as a middleware function/class.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+interface Middleware<Arguments extends Array<any>> {
+  /**
+   * The method that gets called with the same arguments that are passed to the function it is wrapping.
+   *
+   * @param args - Arguments that will be passed to the function - immutable.
+   */
+  call(...args: Arguments): void;
+}
+
+/**
+ * Tells the SDK to apply an array of passed in middleware to the function execution.
+ *
+ * @param middlewares - The middlewares to be executed.
+ */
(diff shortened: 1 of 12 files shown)


src/near-bindgen.ts
```diff
@@ -89,6 +89,50 @@ export function call({
   };
 }
 
+/**
+ * The interface that a middleware has to implement in order to be used as a middleware function/class.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+interface Middleware<Arguments extends Array<any>> {
```
ailisp, line 96: Can we make Middleware a `type = Function`, as we don't have other member than `call`. And let user call it as:
```js
@middleware((args) => near.log(args))
```

tests/__tests__/test-middlewares.ava.js
```diff
@@ -0,0 +1,84 @@
...
+  const args = { id: "test", accountId: "tset" };
+  // Call the function.
+  const result = await ali.callRaw(middlewares, "get_private", "");
+  // Extract the logs.
+  const { logs } = result.result.receipts_outcome[0].outcome;
+  // Create the expected logs.
+  const expectedLogs = [`Log from middleware: ${args}`];
+
+  t.deepEqual(logs, expectedLogs);
+});
```
ailisp, line 84: Great tests that covers compatibility with existing decorators! Could you add a test multiple middleware?

src/near-bindgen.ts
petarvujovic98: @ailisp I can create a commit for this. It's not difficult.

tests/__tests__/test-middlewares.ava.js
petarvujovic98: @ailisp Yes, I will add another method where I pass multiple middlewares.

petarvujovic98:
One thing that I am not satisfied with is type inference in the decorator. It would be ideal if I could get the editor/language server to tell the developer what type the middleware arguments will be when the `middleware` decorator is attached to a method. I have not yet seen something similar, but I will try to come up with a solution.

src/near-bindgen.ts
petarvujovic98: @ailisp This is addressed [here](https://github.com/near/near-sdk-js/pull/263/commits/79a9ecbdb3508e2d753dc852ff880c9bcc865c85).

tests/__tests__/test-middlewares.ava.js
petarvujovic98: @ailisp This is addressed [here](https://github.com/near/near-sdk-js/pull/263/commits/79a9ecbdb3508e2d753dc852ff880c9bcc865c85).

ailisp:
> One thing that I am not satisfied with is type inference in the decorator. It would be ideal if I could get the editor/language server to tell the developer what type the middleware arguments will be when the middleware decorator is attached to a method. I have not yet seen something similar, but I will try to come up with a solution.

This sounds challenging. I think we can address in future PRs. Your current middleware implementation is general, easy-to-use and well tested. IMO, it's ready for merge.

@volovyks Do you observe any issue regarding Petar's implementation?

ailisp (approved):

volovyks (approved):
No objections @ailisp , it is a great implementation.

PR #277, opened 2022-11-02
cli chmod hack deleted

volovyks:


ailisp (commented):
Will the packaged near-sdk-js still installing `near-sdk-js` as executable? Is that because npm/yarn will do that?

volovyks:
@ailisp I checked the docs on this, the only thing we need to do is to add it to bin in package.json. TS CLI file does not need to be 755, and the JS CLI file is already 755 and it stays 755 after we make changes to TS one and `yarn build`.
`chmod` works only on UNIX systems, so it's better if we will not use it.

ailisp (approved):
Looks good!

Issue #280, opened 2022-11-03
Can't use npm package `ajv` and perhaps many others


waynenilsen:
In an attempt to do input validation for our smart contract, as an experiment, I attempted to use the popular `ajv` package.

As a result of [building with TSC with CLI arguments](https://github.com/near/near-sdk-js/blob/eb5f03cf2295554112ad281b909ba588f88c8473/lib/cli/cli.js#L65) **the `tsconfig.json` in a typical location adjacent to `project.json` is ignored**. This is surprising behavior to many and still not well documented by microsoft but unfortunately its true. 

Not having known this fact I added the `allowSyntheticDefaultImports` flag to my `tsconfig.json` but i didn't help! That is when I decided to look at how `near-sdk-js` actually works and discovered this quirk. I was able to reproduce the error in my console and using `npm link ../near-sdk-js` was able to patch it and move forward.

This lead me to another discovery. [`commonjs`](https://github.com/near/near-sdk-js/blob/eb5f03cf2295554112ad281b909ba588f88c8473/lib/cli/cli.js#L76) is commented out however I did need that to allow rollup to build my file.

One third thing that was required by `ajv` and `rollup` interacting was some `rollup-json` plugin. 

---

This is all a very long winded way of perhaps pointing out a deeper issue. `near-sdk-js` is perhaps doing too much OR it may need more flags OR it may need to become smarter.

I think at minimum `commonjs` support should be added to `rollup` not sure why that got removed.

I think, probably, _if_ there is a `tsconfig.json` file present `tsc` should run without arguments. OR there should be a flag to the CLI to avoid compiling typescript (type checking) with `tsc` entirely since `rollup` is doing the transpiling anyway. Those `tsc` checks can be done externally by someone that knows what they're doing.

---

Regarding input validation, since package import is not really working, I went ahead and rolled my own. I am actually very happy with how it came out, extremely lightweight and probably better suited to smart contract programming anyway. But most importantly it worked with the existing build system.

It is also worth mentioning that this was on `0.5.0-1` and I know there has been a release since then and that line i ponted out has been modified with some sort of "don't check libs" flag but I don't think that fixes the issue raised here.

waynenilsen:
now that the build process is broken up this is resolved

PR #283, opened 2022-11-03
feat(cli): new subcommands

waynenilsen:
### Problem Statement

We may want to use parts of this build system with parts of our own build system. I may have use of using my own build script to call rollup and tsc.

### Solution

To start this conversation I suggest we make new subcommands for validating and doing other things step-by-step. This is somewhat designed as a sketch to gather feedback and see if you are interested in this kind of work.

### How has this been tested?

ok I tested out with my contract, this should work fine, will take another look at unit tests

---

[related issue](https://github.com/near/near-sdk-js/issues/280)

### DIFF
--- a/lib/cli/cli.d.ts
+++ b/lib/cli/cli.d.ts
@@ -1,4 +1,16 @@
 #!/usr/bin/env node
+export declare function validateCom(source: string, { verbose }: {
+    verbose: boolean;
+}): Promise<void>;
+export declare function checkTypescriptCom(source: string, { verbose }: {
+    verbose: boolean;
+}): Promise<void>;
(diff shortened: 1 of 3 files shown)


waynenilsen:
another option is to export and/or move the funcitons called in `buildCom` such as `createHeaderFileWithQjsc`

volovyks:
It's a great idea @waynenilsen ! I think we want to have a command for all the steps of the build process + one command that will run the pipeline (`build`). It will help a lot during contract development and library development. Do you want to extend this PR with all the commands?

waynenilsen:
yes i can take care of it

ailisp (approved):
Looks good to me! Yes this is a great help to run one specific step of the build process. Especially for advanced users and near-sdk-js maintainers

PR #285, opened 2022-11-04
cli bash replaced with node

volovyks:


ailisp (approved):
Looks good to me!

src/cli/cli.ts
```diff
@@ -160,7 +162,7 @@ async function createMethodsHeaderFile(rollupTarget: string, verbose = false) {
     ""
   );
 
-  await fs.writeFile(`${buildPath}/methods.h`, methods);
+  fs.writeFileSync(`${buildPath}/methods.h`, methods);
```
ailisp, line 165: What is the reason to prefer writeFileSync over await writeFile?

volovyks: I think they are the same, the only reason is consistency.
The main reason to move from bash to node scripts is Multiplatform (Windows) support.

PR #287, opened 2022-11-07
Example of programmatic update of JS Smart Contract

nsejim:
Tried to implement the programmatic update with JS SDK to be used after locking the contract. 
However this not working. The problem is on how to pass wasm code when calling the call update method and how to get the code in the smart contract.

To test this: 
- cd examples
- yarn
- yarn build:programmatic-update
- yarn test:programmatic-update

The working RUST version is available 
https://github.com/near-examples/update-migrate-rust/blob/main/contracts/self-updates/base/src/update.rs#L10-L28#


examples/__tests__/test-programmatic-update.ava.js
```diff
@@ -0,0 +1,48 @@
...
+test("the contract can update itself", async (t) => {
+  const { ali, programmaticUpdate } = t.context.accounts;
+
+  const code = fs.readFileSync(path.resolve("./build/programmatic-update.wasm"));
+
+  // Ali is the manager, asks the contract to update itself
+
+  // QUESTION: IN WHICH FORMAT SHOULD WE PASS THE WASM CODE AS ARGS? THE CODE IS OF TYPE BUFFER
+  const result = await ali.callRaw(programmaticUpdate, "updateContract", {
+    code
```
ailisp: I think it cannot be in json. As contract is raw bytes, it can contains ascii `"` and ending json string unexpectedly. However, our high level API only takes json serialized contract arguments and this is not customizable.

So, currently it's only possible to use low level APIs to implement your awesome example. Check `test_promise_batch_deploy_call` in tests/src/test_promise_batch_api.js, use `promiseBatchActionDeployContract`, parsing the raw bytes of the as input (passing Uint8Array to callRaw, and that'll be available in contract from `near.input()`)

examples/src/programmatic-update.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  @call({}) // This method update the code
+  updateContract(): void {
+    near.log(`Manager = ${this.manager}`)
+    near.log(`Caller = ${near.predecessorAccountId()}`)
+    assert(near.predecessorAccountId() === this.manager, "Only the manager can update the code")
+    
+    // QUESTION: HOW DO WE GET A WASM CODE PASSED AS ARGS?
+    const args = JSON.parse(near.input());
```
ailisp: Just `near.input()`, without JSON.parse it

examples/src/programmatic-update.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    const promise = near.promiseBatchCreate(near.currentAccountId());
+
+    // QUESTION: DO WE NEED TO RESOLVE THIS PROMISE FOR IT TO RUN?
+    near.promiseBatchActionDeployContract(promise, args.code)
```
ailisp: In the high level API, promise need to be returned to let the runtime handle the promise: `return promise`. In low level API, it's `near.promiseReturn(promise_index)`

examples/__tests__/test-programmatic-update.ava.js
ailisp: The other way is expanding our high level API, more specifically, in near-bindgen.ts, `_getArgs` use hard coded way to deserialize the arg with `JSON.parse`. Making that customizable with option can handle your case specially!

ailisp:
@nsejim Great idea and good experiment on the example! Thank you! The problem is our high level API doesn't yet support this, please see my comment above for details. If you are interested to work on it, it will be great! Or we'll add it to our road map to add it in future.

examples/__tests__/test-programmatic-update.ava.js
nsejim: Thanks. Let me try the above low level API solution based on your reference

examples/src/programmatic-update.ts
nsejim: OK

examples/src/programmatic-update.ts
nsejim: Noted

nsejim:
> @nsejim Great idea and good experiment on the example! Thank you! The problem is our high level API doesn't yet support this, please see my comment above for details. If you are interested to work on it, it will be great! Or we'll add it to our road map to add it in future.

nsejim:
@ailisp Thank you for the comments. I am interested to try to work on it. Putting it on the roadmap is good because it is an important feature to eliminate some convincing arguments to go for RUST.

nsejim:
@ailisp following your suggestion to use the low level APIs, the programmatic update works very well. Thank you.

ailisp:
@nsejim Your code looks good! It's a quite helpful example indeed! Could you add your example to CI here: https://github.com/near/near-sdk-js/blob/develop/.github/workflows/examples.yml ? Thank you!

ailisp (approved):

nsejim:
> @nsejim Your code looks good! It's a quite helpful example indeed! Could you add your example to CI here: https://github.com/near/near-sdk-js/blob/develop/.github/workflows/examples.yml ? Thank you!

Thanks :) Example added to CI

ailisp:
@nsejim seems ci is unhappy. Please `yarn lint` & `yarn format`, then commit & push the changed files. Thanks!

volovyks:
Great contribution @nsejim ! Thank you.

nsejim:
> Great contribution @nsejim ! Thank you.

Welcome. Thank you for the hint!

PR #297, opened 2022-11-11
Update cli.js to avoid error when importing ethers.js

inmarelibero:
this add lets you use the package `ethers.js`, pretty important in some contexts

If rollup is configured in this way, you get the error `Error: 'default' is not exported by node_modules/bn.js/lib/bn.js, imported by node_modules/@ethersproject/bignumber/lib.esm/bignumber.js` when you try to do `import {ethers} from "ethers";` in your smart contract

### DIFF
--- a/lib/cli/cli.js
+++ b/lib/cli/cli.js
@@ -3,6 +3,7 @@ import fs from "fs";
 import path, { basename, dirname } from "path";
 import { nodeResolve } from "@rollup/plugin-node-resolve";
 import sourcemaps from "rollup-plugin-sourcemaps";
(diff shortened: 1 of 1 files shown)


ailisp:
Hi @inmarelibero ! Thanks for your PR! The reason for we not enabling common js import is due to it significantly increased the bundled contract size. If only use ESM packages, only imported parts of packages are included in the generated contract, so allowing commonjs is not suitable for every contract. May you make this as an optional argument to the cli so user can decide to have/not have it?

Also, please update src/cli/cli.ts and run `yarn build`. `lib/cli/cli.js` is a generated file

PR #301, opened 2022-11-15
ABI generation

itegulov:
This is still pretty raw, but sharing for visibility.

Instructions for running:
1. Clone https://github.com/near/near-abi-js to a sibling directory and run `yarn install && yarn build` (needed because it is not released yet)
2. Clone https://github.com/itegulov/typescript-json-schema to a sibling directory and run `yarn install && yarn build` (needed because I had to expose some API that makes life much much easier).
3. Run `yarn build:counter-ts` in `examples` and see what happens :eyes:


src/cli/abi.ts
```diff
@@ -0,0 +1,198 @@
...
+            const message = ts.flattenDiagnosticMessageText(diagnostic.messageText, "\n");
+            if (diagnostic.file) {
+                const { line, character } = diagnostic.file.getLineAndCharacterOfPosition(diagnostic.start!);
+                console.error(`${diagnostic.file.fileName} (${line + 1},${character + 1}): ${message}`);
+            } else {
+                console.error(message);
+            }
+        });
+        return undefined;
+    }
```
itegulov: We can get rid of the preceding CLI step that does typechecking since this piece does effectively the same without having to instantiate `tsc` twice. But up to you guys.

package.json
```diff
@@ -21,6 +21,7 @@
   ],
   "license": "(MIT AND Apache-2.0)",
   "scripts": {
+    "prebuild": "node -p \"'export const LIB_VERSION = ' + JSON.stringify(require('./package.json').version) + ';'\" > src/version.ts",
```
volovyks: Why do we need this script?

package.json
```diff
@@ -46,10 +47,13 @@
     "@rollup/plugin-node-resolve": "^13.1.1",
     "@scure/base": "^1.1.1",
     "commander": "^9.4.1",
+    "near-abi": "file:../near-abi-js",
```
volovyks: @itegulov 
Is it a separate crate in Rust? If yes, looks like we will need to move to the monorepo arch @ailisp.

src/version.ts
```diff
@@ -0,0 +1 @@
+export const LIB_VERSION = "0.6.0";
```
volovyks: Probably we can get it from package.json.

src/cli/abi.ts
volovyks: It is still unclear to me if IBI is a separate package. If it is not - then we can do a TS check once.

package.json
itegulov: It is. Basically an unopinionated set of models that people can use to parse/work with ABIs: https://github.com/near/near-abi-rs. If you don't think it will be useful as its own library, I can move it here for sure.

src/version.ts
itegulov: This relates to what I was talking about in Slack. If I import package.json, then it messes up how sources are placed in the `lib` folder. For transparency this is the Slack message:

>  I figured out why. Apparently if you import any file outside of src, then tsc will maintain the src directory for you. So when I did this:
`import sdkPackage from '../../package.json';`
It made so that package.json was copied to `lib` and hence the source code had to go into `src`. Which makes sense, but came as a bit of surprise.

package.json
itegulov: Basically a roundabout way of getting library version without having to import package.json. Definitely hacky, so feel free to suggest an alternative!

src/cli/abi.ts
itegulov: IBI?

src/version.ts
miraclx: How about `JSON.parse(path.join(__dirname, "..", "package.json"))`?

package.json
ailisp: Since it's similarly organized as rust, as a separate looks good

src/cli/abi.ts
```diff
@@ -0,0 +1,198 @@
+import ts, { Decorator, NodeArray } from "typescript";
+import * as abi from "near-abi";
+import * as TJS from "typescript-json-schema";
+import { JSONSchema7 } from "json-schema";
+import * as fs from "fs";
+import { LIB_VERSION } from "../version.js";
```
ailisp, line 6: It looks like this generated version.js isn't necessary, since you have `parseMetadata` below, which still load package.json?

src/cli/abi.ts
```diff
@@ -0,0 +1,198 @@
...
+    // use built-in default options
+    const compilerOptions = ts.convertCompilerOptionsFromJson(jsonCompilerOptions, basePath).options;
+    const options: ts.CompilerOptions = {
+        noEmit: true,
+        emitDecoratorMetadata: true,
+        experimentalDecorators: true,
+        target: ts.ScriptTarget.ES5,
+        module: ts.ModuleKind.CommonJS,
+        allowUnusedLabels: true,
+    };
```
ailisp: So these overrides options load in basePath/tsconfig.json?

src/cli/abi.ts
```diff
@@ -0,0 +1,198 @@
...
+                                }
+                            });
+                        });
+                    }
+                    if (decoratorName === "view") isView = true;
+                    if (decoratorName === "initialize") isInit = true;
+                });
+                const nearDecoratorsCount = [isCall, isView, isInit].filter((b) => b).length;
+                if (nearDecoratorsCount > 1) {
+                    throw Error("NEAR function cannot be call and view at the same time");
```
ailisp:
```suggestion
                    throw Error("NEAR function cannot be init, call and view at the same time");
```

ailisp:
Generally your implementation and comprehensive validation looks great!

src/cli/abi.ts
```diff
@@ -0,0 +1,198 @@
...
+            if (
+                node.kind === ts.SyntaxKind.MethodDeclaration
+            ) {
+                const methodDeclaration = node as ts.MethodDeclaration;
+                const decorators = methodDeclaration.decorators || ([] as unknown as NodeArray<Decorator>);
+                let isCall = false;
+                let isView = false;
+                let isInit = false;
+                const abiModifiers: abi.AbiFunctionModifier[] = [];
+                decorators.forEach((decorator) => {
```
ailisp, line 102: we may also want to ensure it's under a parent node of `@NearBindgen`

src/cli/abi.ts
itegulov: `parseMetadata` actually loads the package file supplied by the user (i.e. contract's package.json). But I also need to parse the near-sdk-js package.json. We can go with what @miraclx suggested in the other thread though if that looks good to you.

src/cli/abi.ts
itegulov: So I tried to use contract's `tsconfig.json`, but then it picks up all TS files in this project, not just the one you directly specify (e.g. all examples under `/examples`, not just `counter-ts`). I will play around with this a bit more and see what can be done.

src/version.ts
itegulov: @miraclx couldn't use your snippet as `__dirname` is not supported in ES modules. Went with `JSON.parse(fs.readFileSync(path.join(fileURLToPath(import.meta.url), "..", "..", "package.json"), 'utf-8'));` as the next best option.

src/cli/abi.ts
itegulov: I think I figured out the right way to deal with this, PTAL.

While we are at this, how do you think the path to `package.json` and `tsconfig.json` should be specified? I have just added a 3rd and 4th argument to the CLI command, but that's probably not what we want.

itegulov:
I am working on stabilizing this and I have a couple of things that I am not sure how to go about.

1. Should this be an opt-in feature for now? How do you guys imagine the CLI interface for ABI generation should look like? Do we also provide a separate command `abi` that just generates ABI without building the contract? I can match what we have in Rust, but not sure if that makes sense as the build tools are quite different, so your call.
2. I have released https://github.com/itegulov/typescript-json-schema as `near-typescript-json-schema` for now just to unblock this PR, but not sure how you would like to handle this fork. We have a similar situation with https://github.com/near/schemafy/, so we just moved the fork under `near` org and are maintaining it ourselves. Ideally we should discuss this with the original repo maintainers and see if they would be open to exposing the APIs that we need, but in my experience open-source work is pretty slow, so I wouldn't count on it short-term (we are still waiting for schemafy maintainers' response for instance).

So far most examples work, only NFT is failing because `bigint` is apparently not supported by `typescript-json-schema`. There is [this PR](https://github.com/YousefED/typescript-json-schema/pull/433) that was never merged though, so it seems like I can easily add support in our fork if you don't have any objections.

src/cli/abi.ts
volovyks: Sorry, "ABI"

volovyks:
@itegulov our build command now consists of several steps. I think ABI can become one of them. It should be easy to have as a separate command as well. I do not think that we need to keep it as a separate project.

src/version.ts
miraclx: PS: this is equivalent to

`JSON.parse(fs.readFileSync(fileURLToPath(new URL("../package.json", import.meta.url)), "utf-8"))`. 😄

itegulov: This wouldn't work on Windows though, right?

miraclx: Tested, it works. `fileUrlToPath` makes the conversion.

itegulov: Ah, nice! I will change this over then

examples/src/parking-lot.ts
```diff
@@ -67,7 +67,7 @@ export class ParkingLot {
   }
 
   @view({})
-  getCarSpecs({ name }: { name: string }) {
+  getCarSpecs({ name }: { name: string }): CarSpecs {
```
ailisp, line 70: will we get warnings (or configure tsconfig in a way) when method doesn't have return type?

ailisp (approved):
Nice work! It's very solid implementation! Also thank you for fixing a few signatures!

volovyks (approved):
Great work @itegulov!
Merging now, but do you think it's possible to test ABI somehow?

PR #304, opened 2022-11-17
Use TS ? to replace option in nft standard interface

### LINKED ISSUE #289: Use ts style option instead of `?` in NFT standard
From @volovyks :
> I'm curious if we are using Rust style Option here because it's the only way to make it work the same way as in Rust standard. Can it work with ? somehow?

ailisp:
Fix #289


ailisp:
@volovyks Ready for review. workspace starts to hit "key not found" error again, and pass on retrying. Not sure it worth investigaing.

near-contract-standards/src/non_fungible_token/metadata.ts
```diff
@@ -9,29 +8,29 @@ export class NFTContractMetadata {
...
 
   constructor() {
     this.spec = NFT_METADATA_SPEC;
     this.name = "";
     this.symbol = "";
-    this.icon = null;
-    this.base_uri = null;
-    this.reference = null;
-    this.reference_hash = null;
+    this.icon = undefined;
```
volovyks: I think there is no point to set it to undefined. It's undefined by default.

near-contract-standards/src/non_fungible_token/metadata.ts
```diff
@@ -45,10 +44,10 @@ export class NFTContractMetadata {
   assert_valid() {
     assert(this.spec == NFT_METADATA_SPEC, "Spec is not NFT metadata");
     assert(
-      (this.reference != null) == (this.reference_hash != null),
+      (this.reference !== undefined) == (this.reference_hash !== undefined),
```
volovyks: What will happen if it is null? Should we check for it?

near-contract-standards/src/non_fungible_token/metadata.ts
```diff
@@ -45,10 +44,10 @@ export class NFTContractMetadata {
...
       "Reference and reference hash must be present"
     );
-    if (this.reference_hash != null) {
+    if (this.reference_hash !== undefined) {
```
volovyks: Same here and in other places.

volovyks (commented):
So, what is the behavior in Rust?
If the field is Optional and None, how it will be represented in a view function? Should we strive for the same behavior here?

near-contract-standards/src/non_fungible_token/metadata.ts
ailisp: Good point! In rust, json with these field are returned, with null as value. so I'll change it to null.

volovyks (approved):
Ok, I think my concerns about null vs undefined should be ok because we are initializing everything with null in the constructor.

PR #308, opened 2022-11-22
Use uint8array

ailisp:
This PR will fix the issues mentioned in #117 and #195.

- [x] Implement #314 
- [x] Breaking changes: Redesign all relevant APIs to take Uint8Array instead of Bytes. Drop Bytes.
- [x] Redesign all collections, take string as keys. string can now be any UTF-8 string not only latin1 strings. And all collection APIs remain backward-compatible
- [x] Low level API tests, bytes tests and collection tests work
- [x] Fix all tests
- [x] Add tests to cover all cases conversion/encoding between string and Uint8Array
```
- any (utf8 seq, non utf8 seq) uint8array to string latin1
- any latin1 string (utf8 seq, non utf8 seq) to uint8array
- any (utf8 seq, non utf8 seq) uint8array to string utf8
- any utf8 string (utf8 chars, latin1 chars that happens to be utf8 seq, latin1 chars) to uint8array
- collections ( JSON.stringify state ), JSON.stringify utf8 input
```
- [x] Fix all examples


ailisp:
run ci tasks locally pass, mark it ready for review. @aesilevich to review changes in C code

examples/__tests__/test-cross-contract-call.ava.js
```diff
@@ -52,12 +52,13 @@ test("Person can be set on-call if AVAILABLE", async (t) => {
   // Ali set her status as AVAILABLE
   await ali.call(statusMessage, "set_status", { message: "AVAILABLE" });
   // Bob sets Ali on-call
-  await bob.call(
+  let r = await bob.callRaw(
```
volovyks: Looks like you forget to delete changes in this file.

packages/near-sdk-js/src/api.ts
```diff
@@ -48,86 +48,97 @@ interface Env {
...
+    methodName: string,
+    args: Uint8Array,
     amount: NearAmount,
     gas: NearAmount,
     weight: GasWeight
   ): void;
   promise_results_count(): bigint;
   promise_result(promiseIndex: bigint, register: Register): PromiseResult;
   promise_return(promiseIndex: bigint): void;
+
```
volovyks, line 145: Let's add a comment that will separate our host functions from helper C functions.

packages/near-sdk-js/src/api.ts
```diff
@@ -177,17 +188,17 @@ export function signerAccountPk(): Bytes {
...
   env.predecessor_account_id(0);
-  return env.read_register(0);
+  return str(env.read_register(0));
 }
 
 /**
  * Returns the account ID of the current contract - the contract that is being executed.
  */
-export function currentAccountId(): Bytes {
+export function currentAccountId(): string {
```
volovyks, line 208: Users will appreciate this.

packages/near-sdk-js/src/api.ts
```diff
@@ -7,21 +7,21 @@ const EVICTED_REGISTER = U64_MAX - 1n;
...
+  read_register(register: Register): Uint8Array;
 
   // Storage
-  storage_read(key: Bytes, register: Register): bigint;
-  storage_has_key(key: Bytes): bigint;
-  storage_write(key: Bytes, value: Bytes, register: Register): bigint;
-  storage_remove(key: Bytes, register: Register): bigint;
+  storage_read(key: Uint8Array, register: Register): bigint;
+  storage_has_key(key: Uint8Array): bigint;
+  storage_write(key: Uint8Array, value: Uint8Array, register: Register): bigint;
```
volovyks, line 31: We talked about it, but again, what defines if we have string or Uint8Array here? And is it possible to do the conversion in api.ts so users can use strings? I guess they expect "key" to be a string.

packages/near-sdk-js/src/collections/lookup-map.ts
```diff
@@ -13,16 +13,16 @@ export class LookupMap<DataType> {
...
   /**
    * Checks whether the collection contains the value.
    *
    * @param key - The value for which to check the presence.
    */
-  containsKey(key: Bytes): boolean {
+  containsKey(key: string): boolean {
     const storageKey = this.keyPrefix + key;
-    return near.storageHasKey(storageKey);
+    return near.storageHasKey(encode(storageKey));
```
volovyks: I'm talking about places like this. Can we move this conversion to the API layer?

packages/near-sdk-js/src/near-bindgen.ts
```diff
@@ -166,29 +166,31 @@ export function NearBindgen({
...
 
       static _getState(): unknown | null {
-        const rawState = near.storageRead("STATE");
+        const rawState = near.storageRead(bytes("STATE"));
         return rawState ? this._deserialize(rawState) : null;
       }
 
       static _saveToStorage(objectToSave: unknown): void {
-        near.storageWrite("STATE", this._serialize(objectToSave));
+        near.storageWrite(bytes("STATE"), this._serialize(objectToSave));
```
volovyks: Same comment again. Can be unintuitive for devs. What will happen if we omit `bytes()` call?

tests/__tests__/test_highlevel_promise.ava.js
```diff
@@ -168,6 +168,9 @@ test("cross contract call success then call a panic method", async (t) => {
     }
   );
   // the last promise fail, cause the transaction fail
+  console.log(
```
volovyks: Probably should not be added.

tests/__tests__/test_promise_api.ava.js
```diff
@@ -232,8 +232,10 @@ test("promise batch deploy contract and call", async (t) => {
     caller2Contract,
     "test_promise_batch_deploy_call",
     "",
-    { gas: "200 Tgas" }
+    { gas: "300 Tgas" }
   );
+  console.log(JSON.stringify(r, null, 2));
```
volovyks: One more ;)

tests/src/highlevel-promise.js
```diff
@@ -120,7 +120,7 @@ export class HighlevelPromiseContract {
     return {
       ...callingData(),
       promiseResults: arrayN(near.promiseResultsCount()).map((i) =>
-        near.promiseResult(i)
+        str(near.promiseResult(i))
```
volovyks: Any use cases we want to represent `promiseResult` as something else but string?
(this one is about moving conversion to the API level again)

tests/yarn-error.log
```diff
@@ -0,0 +1,2205 @@
+Arguments: 
+  /home/bo/.nvm/versions/node/v16.14.0/bin/node /home/bo/.nvm/versions/node/v16.14.0/bin/yarn add near-sdk-js
```
volovyks: Let's get rid of this file.

examples/src/clean-state.js
```diff
@@ -1,19 +1,23 @@
-import { NearBindgen, call, view, near } from "near-sdk-js";
+import { NearBindgen, call, view, near, encode, decode } from "near-sdk-js";
 
 @NearBindgen({})
 export class CleanState {
   @call({})
   clean({ keys }) {
-    keys.forEach((key) => near.storageRemove(key));
+    keys.forEach((key) => near.storageRemove(encode(key)));
```
volovyks: Developers should not do encoding on the contract level.

examples/yarn-error.log
```diff
@@ -0,0 +1,2269 @@
+Arguments: 
+  /home/bo/.nvm/versions/node/v16.14.0/bin/node /home/bo/.nvm/versions/node/v16.14.0/bin/yarn
```
volovyks: Let's delete this file.

volovyks (approved):
Great work @ailisp ! Huge effort.
My main suggestion is to move all the conversions to the API level. I left a few comments about that. But we can discuss it separately, feel free to merge this.

packages/near-sdk-js/src/api.ts
ailisp: If these functions also doing the high level task: decode / encode string into Uint8Array, then user do not have an API to deal with raw bytes. At current form they can compose high level functionality: write string to state by combining storage_write + bytes/encode

> what defines if we have string or Uint8Array here

For apis like `current_account_id()`, we are sure it returned is a string. For apis like `storage_read`, it takes and returns array with any bytes, which is more fit Uint8Array.

> And is it possible to do the conversion in api.ts so users can use strings?

Yes. Discussed further in your next comment

tests/src/highlevel-promise.js
ailisp: One example would be promise return one's public key.  And callback function take that public key to do something. Public keys that host function returns and takes are binary format.

packages/near-sdk-js/src/collections/lookup-map.ts
ailisp: What do you think of this: have two APIs, `storageHasKey` and `storageHasKeyRaw`. where `storageHasKey(k) = storageHasKeyRaw(encode(k))`. The benefit is backward compatible and in most case the string-version it's what user want. And they can still achieve raw bytes manipulation with raw-version. 

The disadvantage is this API might hide the fact that state is in the form of binary. When attempt to use the Low level API, it is expected that they already know nomicon. And to me it's unexpected that `near.storageHasKey` is not `env.storage_has_key` but `near.storageHasKeyRaw` is. This makes api naming inconsistent. For example, `input` still returns Uint8Array, should it be `inputRaw` and `input`?

examples/src/clean-state.js
ailisp: Same issue as above: https://github.com/near/near-sdk-js/pull/308#discussion_r1045451673

packages/near-sdk-js/src/near-bindgen.ts
ailisp: In ts, compile error. In js, runtime error, smart contract panicked, expect Uint8Array for key.

tests/src/highlevel-promise.js
volovyks: So, if they both have string in their API, it will work? But we will do the conversion 2 times (gas issue)?

packages/near-sdk-js/src/api.ts
volovyks: Actually user has it, because they can use `env.function()` without API wrapper function. But it's less convinient, I agree.

Maybe we can separate this into 2 issues. Function params and return types. Function params can accept `string | Uint8Array` and return types can be Uint8Array if really necessary. `functionRaw` and 'function' is also an option.

It looks like a purely DevEx problem, let's get input from DevRel team.

packages/near-sdk-js/src/collections/lookup-map.ts
volovyks: Hm... maybe not raw, maybe inputBytes() ?
(more comments above)

packages/near-sdk-js/src/api.ts
idea404: I feel the `function` and `functionRaw` distinction makes sense. At the same time, haven't seen devs using these functions or asking about them often, so implementing two functions without them being popular feels contradictory. In any case, having a function that accepts a single param type is closer to single resp principle.

tests/src/highlevel-promise.js
ailisp: Yes. conversion 2 times adds unnecessary gas consumption

packages/near-sdk-js/src/collections/lookup-map.ts
ailisp: inputBytes sounds good to me

packages/near-sdk-js/src/api.ts
ailisp: supporting points:
> ... it's less convinient, I agree.
> functionRaw and 'function' is also an option.
> function and functionRaw distinction makes sense
> having a function that accepts a single param type is closer to single resp principle.

about alternative:
> Actually user has it, because they can use env.function() without API wrapper function

Right now `env` is not exported, and we need more discussion on whether to export it (it doubles host function APIs)

>  Function params can accept string | Uint8Array and return types can be Uint8Array if really necessary

I thinked about it, will still be breaking changes for `function(string | Uint8Array) -> Uint8Array` because return type was string, and if return `string | Uint8Array` based on param type, it's kind of magic.

> At the same time, haven't seen devs using these functions or asking about them often,

Yeah it is not often, I've seen one ask: https://github.com/near/near-sdk-js/pull/195#issuecomment-1293070576

Combining your thoughts, I'll go ahead with ` function` + `functionRaw` (I assume Serhii is okay with both functionRaw and functionBytes and @idea404 prefer functionRaw, and naming of less popular low level API can be changed or alias later)

ailisp:
Based on above discussion, I make storage / valueReturn / input / promise API with two versions. One is string version. It works with any string and utf-8 decode the string to fit NEAR's binary state / arguments interface. The other one is raw version that works directly on `Uint8Array` and developer can handle customized encoding. 

Note that even though string version has the same name, same interface as before, it is not fully backward compatible with original APIs. Take storageWrite as example, the comparison is:
- If string is fully ascii, the behavior is the same. 
- If string contains charCode > 255 unicode char, original API doesn't work, current string API works as expected
- If string contains 127 < charCode <= 255, original API stores exactly one byte per char, current string API treat charCode as a subset of unicode char and encode it in utf-8, resulting use two bytes to store one char. It is not as storage efficient as the raw version (storageWriteRaw) or the original version, but result is guaranteed to be correct when read back.

PR #309, opened 2022-11-23
Include Example and Test for Promises Created in For Loops

idea404:
This PR includes a failing example for promises created in for loops


idea404:
I believe the CI tests the current code against the existing `package.json`? Which would be why tests are passing here? 

In my local env this is the test output: 

image

@petarvujovic98 also, any clue why the terminal output is printed line by line now like this when building?

gagdiez:
@idea404 `promise.then` and `promise.and` do not append to the variable promise, but return a new NearPromise, this means that:

```ts
const promise = NearPromise.new(account)
promise.functionCall(something) /// here you are missing the returned object
return promise // you are returning the result of doing `NearPromise.new()` which is 0 apparently.
```

What you meant to do is:

```ts
let promise = NearPromise.new(contracts[0]).functionCall()

for i in 1...3 {
   promise = promise.and( NearPromise.new(contracts[i]).functionCall() )
}

promise = promise.then( callback )

return promise.asReturn()
```

Also, in the callback you are not currently returning the counter value.

Applying all those changes make the tests pass.

idea404:
@gagdiez Perfect! Tests now do pass. I'll leave merging this PR here as an example of how to loop promises to one another to the discretion of the maintainers.

gagdiez:
one last thing I would suggest is removing most logs and the variable `callCount`, in the callback you can simple pass or use the `array length` to know the number of calls.

petarvujovic98:
@idea404 I see that the CLI has been updated with new commands, and in doing so there was a degression where instead of creating a new `signale` instance for each command they are using a global one. This is the [commit](https://github.com/near/near-sdk-js/commit/3f170ecaeb55e7fd877b344f18d7fd0348496f80) that changed this - line 46 to be exact. The solution for this would be to simply create a new `signale` instance for each command function with the interactive option configured the same way as it was in the build command, and of course changing the scope to be the command name and to actually use those instances in logging instead of the global variable.

petarvujovic98:
I've addressed the logging issue in #310 .

ailisp (approved):
Very good example! Thanks @idea404 and @petarvujovic98

ailisp:
Oh @idea404 you need to add the test to ci, in .github/workflows/examples.yml

volovyks:
Thank you @idea404 !

PR #313, opened 2022-12-01
Exports refactored

volovyks:
Let's follow this standard for exported functions and primitives. The file where it's defined is the only source of truth to check if it's exported. In case we need internal utils or something similar, they should be named accordingly and not be exported at all.


ailisp (approved):
Nice!

PR #319, opened 2022-12-04
Extending contract

### LINKED ISSUE #290: Helper class inheritance or code generation to quickly build standard contract

volovyks:
Let's define if we want our contract to be extendable. Looks good, but it can lead to some issues with code generation (@waynenilsen ) and ABI (cc @itegulov).
@MaximusHaximus what is your opinion on this? Previously your advice was to avoid class extensions. But contract composability is a cool feature to have.

Note 1: now it does not work, functions from the parent class are not generated or exported.

Note 2: In Rust composability is achieved with macros: https://github.com/near/near-sdk-rs/blob/master/examples/non-fungible-token/nft/src/lib.rs#L103
Not something users want I think, especially in JS projects.

cc @ailisp


examples/__tests__/test-counter.ava.js
```diff
@@ -8,14 +8,30 @@ test.beforeEach(async (t) => {
...
+
   // Deploy the counter contract.
-  const counter = await root.devDeploy(
-    process.env["COUNTER_LOWLEVEL"]
-      ? "./build/counter-lowlevel.wasm"
-      : process.env["COUNTER_TS"]
-      ? "./build/counter-ts.wasm"
-      : "./build/counter.wasm"
-  );
+  const counter = await root.devDeploy(counterContract);
```
ailisp, line 35: Adding a test for `reset`?

ailisp: oh I didn't see this:

> Note 1: now it does not work, functions from the parent class are not generated or exported.

NearBindgen doesn't walk through parent class, it may be possible to identify extends pattern in our babel pass then recursively NearBindgen on parent class

volovyks: Yes, it should be possible (not tested), the question is - is it the best approach? I will experiment with this.

volovyks:
Closing in favor of https://github.com/near/near-sdk-js/pull/325

PR #321, opened 2022-12-04
NFT Implementation simplification

### LINKED ISSUE #290: Helper class inheritance or code generation to quickly build standard contract

volovyks:
@ailisp deleted redundant interfaces (tell me, if we need them for some reason).

### DIFF
--- a/examples/src/standard-nft/my-nft.ts
+++ b/examples/src/standard-nft/my-nft.ts
@@ -12,7 +12,6 @@ import {
 } from "near-sdk-js";
 import {
   NFTContractMetadata,
-  NonFungibleTokenMetadataProvider,
   TokenMetadata,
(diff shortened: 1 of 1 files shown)


ailisp (commented):
It's a code style problem, I think both your approach and mine works. 

In rust, these trait are required and impl of the traits appear outside `impl MyNFT`. In ts it appears together in the class body, so `implements XXXInterface` is optional. However, I think it would be nice to explicitly write these interfaces so if you forget implement any of the method, e.g. `NonFungibleTokenEnumeration.nft_total_supply()`, tsc and vscode will gives you an error:
image

volovyks:
Ok @ailisp ,  close it.
It will not be an issue if we will find a good way for composability.

PR #322, opened 2022-12-05
fix: ensure that NEAR functions have explicit types

itegulov:
ditto


volovyks (approved):
Thank you @itegulov !

PR #323, opened 2022-12-06
Monorepo implemented with Turborepo

### LINKED ISSUE #291: Best way to organize near-contract-standards package
A few options: as a part of near-sdk-js, a separate npm package, a separate repo or turn this repo into a workspace

volovyks:
- Project refactored to monorepo
- Now standards are a separate project
- Turborepo build system is used to orchestrate all the packages
- All `link near-sdk-js` and `cd examples` hacks were removed from build scripts
- Imports organized
- linting added to `cli/` subproject

Such architecture will allow us to scale the project in the future, separate `near-sdk-js` to multiple packages, use 'caching', etc.


ailisp (approved):
Awesome! This is very good project setup. Just one thing: examples are removed from ci, is it intentional?

volovyks:
@ailisp now, when you are running `pnpm build` from the root folder, `Turbo` will execute this command in each package (if it's available) in proper order: near-sdk-js, then standards, then examples and tests since they depend on the first two.
When you run `pnpm test` it will execute tests from all packages in parallel.

With this setup it is faster and more intuitive to have 1 CI job. Linting is also a part of it now. Check the steps and logs.

PR #331, opened 2023-01-11
FT

volovyks:


packages/near-contract-standards/src/fungible_token/events.ts
```diff
@@ -0,0 +1,223 @@
...
+function new_141(version: string, event_kind: Nep141EventKind): NearEvent {
+    return new Nep141Event(version, event_kind);
+}
+
+function new_141_v1(event_kind: Nep141EventKind) : NearEvent {
+    return new_141("1.0.0", event_kind)
+}
+
+// #[cfg(test)]
+// mod tests {
```
ailisp: Remove these?

packages/near-contract-standards/tsconfig.json
```diff
@@ -24,7 +24,8 @@
   "files": [
     "src/index.ts",
     "src/non_fungible_token/approval/approval_receiver.ts",
-    "src/non_fungible_token/core/receiver.ts"
+    "src/non_fungible_token/core/receiver.ts",
+    "src/fungible_token/core_impl.ts"
```
ailisp: If everything is imported in index.ts, `src/index.ts` would be sufficient.

packages/near-contract-standards/src/index.ts
```diff
@@ -1,2 +1,4 @@
-/** Non-fungible tokens as described in [by the spec](https://nomicon.io/Standards/NonFungibleToken/README.html). */
+/** Non-fungible tokens as described in [by the spec](https://nomicon.io/Standards/NonFungibleToken). */
```
ailisp, line 1: Nice catch!

packages/near-contract-standards/src/fungible_token/core.ts
```diff
@@ -0,0 +1,55 @@
...
+     * This method must to be able to accept attached deposits, and must not panic on attached deposit.
+     * Exactly 1 yoctoNEAR must be attached.
+     * See [the Security section](https://github.com/near/NEPs/issues/141#user-content-security) of the standard.
+     *
+     * Arguments:
+     * @param receiver_id - the account ID of the receiver.
+     * @param amount - the amount of tokens to transfer. Must be a positive number in decimal string representation.
+     * @param memo - an optional string field in a free form to associate a memo with this transfer.
+     */
+    ft_transfer(receiver_id: AccountId, amount: Balance, memo: Option<String>);
```
ailisp: These probably should be:
```
    ft_transfer({receiver_id, amount, memo}: {receiver_id: AccountId, amount: Balance, memo: Option<String>});

```

Because Rust ft contract can be called with JSON args (tested with near-cli), in JS contract this is achieved by passing single args,

ailisp: Same for other `ft_*` methods

packages/near-contract-standards/src/fungible_token/core_impl.ts
```diff
@@ -0,0 +1,342 @@
...
+    constructor(prefix: IntoStorageKey) {
+        const storage_prefix = prefix.into_storage_key();
+        this.accounts = new LookupMap<Balance>(storage_prefix);
+        this.total_supply = 0n;
+        this.account_storage_usage = 0n;
+        this.measure_account_storage_usage();
+    }
+
+
+    @call({})
```
ailisp: Probably should not marked with `@call` and `@view`? since this is not a `@NearBindgen` contract

ailisp (commented):
The architecture and implementation looks great! I think a contract using this ft library and tests are necessary to ensure this works. Something like examples/src/standard-nft/my-nft.ts (for contract) and examples/__tests__/standard-nft (for tests).

packages/near-contract-standards/src/fungible_token/core_impl.ts
volovyks: Yes, agree. It will not work right now. But in the futer, if we will make such contracts extendable, we will need to have @NearBindgen in FT and NFT implementations.

ailisp: Ah got it! I still suggest to remove it for now, as NFT doesn't have this and NF has it will cause confusion. Let's add it in the next PR that makes contracts extendable

ailisp:
Looks mostly good to me! It's a big effort. I see you have implemented the ft using standard example (my-ft). The only thing missing is some tests to my-ft, let's just port this one: https://github.com/near/near-sdk-rs/blob/master/examples/fungible-token/tests/workspaces.rs

volovyks:
@ailisp 
yes,that is what I'm doing :)

ailisp:
Great progress in this huge-effort task!

`ft-tests.js` is not runing in the CI, need to rename to `ft-tests.ava.js`?

examples/__tests__/standard-ft/ft-tests.ava.js
```diff
@@ -0,0 +1,291 @@
...
+    //     ValueOrReceiptId::Value(val) => {
+    //         let bytes = base64::decode(&val)?;
+    //         let used_amount = serde_json::from_slice::<U128>(&bytes)?;
+    //         assert_eq!(used_amount, TRANSFER_AMOUNT);
+    //     }
+    //     _ => panic!("Unexpected receipt id"),
+    // }
+    // assert!(res.json::<bool>()?);
+
+    const res = await ftContract.view("ft_total_supply", {});
```
ailisp: TODO transfer_call in this test case?

examples/__tests__/standard-ft/ft-tests.ava.js
```diff
@@ -0,0 +1,291 @@
...
+    await ftContract.call(
+        ftContract,
+        "init_with_default_meta",
+        {
+            owner_id: ftContract.accountId,
+            total_supply: INITIAL_BALANCE
+        }
+    );
+
+    const defiContract = await root.devDeploy("./res/defi.wasm");
```
ailisp, line 35: What is the defi contract? Could you briefly document its purpose and interface here?

ailisp (approved):
Looks like all tests are fixed. The implementation looks solid. There's two comments about some small clean up, which would be nice to address. But generally this is good enough, feel free to merge it!

PR #332, opened 2023-01-17
0.7 release

ailisp:
0.7 pre-release has tested by me and @idea404 . Want to get this merged before #331 , because FT will happen after near-contract-standards bump to 0.7. Shouldn't cause conflictions.


idea404:
I'm getting this error when I try building a contract with this new version: 

image

For these repos: 

https://github.com/near-examples/donation-js
https://github.com/near-examples/coin-flip-js
https://github.com/near-examples/counter-js

volovyks:
@idea404 looks like it's related to the ABI. It's not part of the build process.
@itegulov any idea what is the problem here?

volovyks (approved):
I have no concerns other that the one mentioned by @idea404

ailisp:
@idea404 Thanks for the thorough testing! I'll make ABI generation turn off by default and try again.

itegulov:
Hmm, so it seems like ABI logic is failing on this line: https://github.com/near/near-sdk-js/blob/develop/packages/near-sdk-js/lib/cli/abi.js#L53. So your tsconfig is technically not a valid JSON file because of this trailing comma [here](https://github.com/near-examples/donation-js/blob/master/contract/tsconfig.json#L9).

But it seems like trailing commas are generally supported for tsconfig. @ailisp @volovyks do you guys know what would be the canonical way to parse such JSON?

ailisp:
@itegulov Nice catch! I realized the same issue as you commented. I searched around https://github.com/json5/json5 seems a popular solution. And it's adding a dependency to the CLI, not the resulting contract, should be okay.

itegulov:
Ok, I can prepare a PR later today if you aren't taking this yourself

UPD: nevermind, I see that you are on it already 😅

ailisp:
donation-js works with 0.7.0-4 [71112b7](https://github.com/near/near-sdk-js/pull/332/commits/71112b75813fe4f18ccd17b9778ddb81fcaaf5cb) with a bit fix, in contract and in config file. The working contract: https://github.com/ailisp/donation-js/tree/use-near-sdk-js-0.7

But https://github.com/near-examples/coin-flip-js still has trouble to generate abi, the problems seems to be generate abi with this line:

```
type Side = 'heads' | 'tails'
```

error:
```
[generateAbi] › …  awaiting  Generating ABI...
/home/bo/workspace/coin-flip-js/contract/node_modules/.pnpm/near-sdk-js@0.7.0-4/node_modules/near-sdk-js/lib/collections/unordered-map.d.ts (50,5): A computed property name in an ambient context must refer to an expression whose type is a literal type or a 'unique symbol' type.
```

So for this release I'll make abi generation optional and let's figure out this case in a separate issue!

ailisp:
Also modify https://github.com/ailisp/coin-flip-js/tree/use-near-sdk-js-0.7 and https://github.com/ailisp/counter-js/tree/use-near-sdk-js-0.7 to build with 0.7. @idea404 let me know if you have other concerns! Note current latest release is 0.7.0-5.

gagdiez:
@ailisp @volovyks I was testing our hello world example [1], and I am getting an error (see image bellow). To reproduce it, simply open the link [1], change the version inside the `./contract/package.json` to `"near-sdk-js": "0.7.0-5"`.

[1] https://gitpod.io/#/https://github.com/near-examples/hello-near-js


Screenshot 2023-01-19 at 11 28 39

ailisp:
@gagdiez Thanks for testing! Remove the contract/babel.config.json and it builds. Actually I found every 0.6.0 contract need this step to build on 0.7.0, because 0.7.0 builtin the babel config and have a different path to look for our custom babel plugin (near-bingen-exporter, in above error message).

I think this migration step worth mention in readme

Issue #302, opened 2022-11-16
UnorderedMap And Unordered Set Should Have a Way to Return Keys


doriancrutcher:
I was writing an example and realized that an UnorderedMap and an Lookup Map should have a way to return all the keys within the collection. 

In my example I had to store the keys separately in an UnorderedSet to be able to. accomplish this which seems like a redundant use of storage.

ailisp:
Currently, you can use this way to retrieve all keys of your UnorderedMap:
```js
    for (let i = 0; i < m.length; i++) {
      const k = m.keys.get(i);
// k is the key.
    }
```

volovyks:
@doriancrutcher do you think this solution is good enough? Or do you want to have a separate function for that? And is there any diff compared to the Rust version?

ailisp:
Suggested API is added in #337

PR #337, opened 2023-02-01
Enumerate keys/elements and pagination

ailisp:
Address https://github.com/near/near-sdk-js/issues/302


ailisp:
weird, the build and nft standard tests are passing locally but fail on ci.

Update: the problem is turbo incorrectly cache the near-contract-standards build locally. I'll create an issue

volovyks (approved):
Nice work @ailisp !

PR #340, opened 2023-02-03
Fix build script when there is space in project path

### LINKED ISSUE #248: Build fails when there's a space in directory name
Upon building a javascript smart contract the build fails with the error message 

`NO such file or directory\n` 

If the name of the directory contains a space 

in this case my directory name was `Office Hours` instead of `Office_Hours`

ailisp:
Fix #248 , main problem is fixed in near/binaryen wasi-stub wrapper script, where it didn't handle space in path


volovyks (approved):
Cool

PR #341, opened 2023-02-07
gas and storage benchmarks

ailisp:
Add benchmark, readme and tips to benchmark your use case.


pnpm-workspace.yaml
```diff
@@ -2,3 +2,4 @@ packages:
   - "examples"
   - "tests"
   - "packages/*"
+  - "bench"
```
volovyks: nit

volovyks: "Bench" gives me wrong associations. Maybe "benchmark" is better?

bench/src/lowlevel-minimal.js
```diff
@@ -0,0 +1,4 @@
+import { near } from "near-sdk-js";
+
+export function empty() {  
+}
```
volovyks: nit

volovyks: + lots of new empty lines at the end of files

volovyks (approved):
Looks great overall! My main concern is that it's better to store benchmark results in separate autogenerated files. It will allow us to update them with time, since gas usage will be constantly changing. But that can be done later, not a blocker now.

Issue #121, opened 2022-06-24
Standalone contract function name conflicting issue


ailisp:
If user write the contract method name same as one of C function, it would case build or execute fail, depending on which is conflict. Need to figure out a way to address this kind of confliction

osalkanovic:
@ailisp Maybe one of solutions can be to write script to load raw` builder.c ` and to match functions with regex, after that to check before build does any method from contract include matched functions from builder

ailisp:
Good idea!

ailisp:
Fixed in #346

PR #346, opened 2023-02-23
Error and stop build a contract that has method name confliction

ailisp:
This PR is to address #121.

The near-sdk-js builds a contract by generate a C file and turns each contract method to a C function, then compile this C file into a wasm file. Therefore, it's possible the contract method name conflicts with the C function name in `builder.c`. 

Because a contract method name `foo` will generate a `void foo(void)` in the C file, there're totally 4 possible cases of different result will happen. Those are a contract method name `foo` becomes the same as: 
1. a `void foo(void)` in `builder.c` or it's include
2. an `extern void foo(void)` in `builder.c`or it's include
3. a `foo` function with other signature in `builder.c` or it's include. E.g.: `strcpy`
4. an `extern foo` function with other signature. E.g.: `storage_read`.

Case 1. doesn't exist as we checked `builder.c` and it's include. Case 2. only has one function `panic`. If user attempts to define a `panic` contract method, it will cause `env.panic()` call that method instead, which causes misleading result. In this PR this becomes an error. In case 3 and 4, the `C -> WASM` step C compiler gives error because breaks the C signature. In this PR, it collects the error message, transform it into a user friendly one and inform user to avoid using `foo` and ends the build.

All cases are tested with the CLI in this PR. It reports errors like this:
- case 2:
image
- case 3&4:
image

### DIFF
--- a/packages/near-sdk-js/lib/cli/cli.js
+++ b/packages/near-sdk-js/lib/cli/cli.js
@@ -206,6 +206,10 @@ async function createMethodsHeaderFile(rollupTarget, verbose = false) {
     }
     const mod = await import(`${PROJECT_DIR}/${rollupTarget}`);
     const exportNames = Object.keys(mod);
+    if (exportNames.includes('panic')) {
+        signal.error("'panic' is a reserved word, please use another name for contract method");
+        process.exit(1);
+    }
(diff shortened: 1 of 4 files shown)


volovyks (approved):
Good explanation and simple solution, nice work!

PR #352, opened 2023-03-03
Migration example

### LINKED ISSUE #265: Contract state migration pattern
Currently there is no way to prevent the `getState` function call on smart contracts in JavaScript. For our Rust SDK we have the [ignore_state](https://github.com/near/near-sdk-rs/blob/cf25ec2fec72ae9531262250ab8984c7bf7a6d86/near-sdk-macros/src/core_impl/info_extractor/init_attr.rs) `init` attribute.

We should handle this before v1.0 in my opinion.

My initial proposal would be to include an additional parameter to the `call` decorator, as this is the type of function that a migration function would be, and check for the existence of it in the `near-bindgen-exporter` step to see if we should or should not run a `getState` function call for that function.

volovyks:
It appeared that additional parameter does not play well with the current design.
I've added new @migrate decorator that is similar to @initialize, but does not read or checks the state.


examples/src/state-migration.ts
```diff
@@ -0,0 +1,70 @@
...
+    countMessages(): number {
+        return this.messages.toArray().length;
+    }
+
+
+    @migrate({})
+    migrateState(): Vector<NewMessageFormat> {
+        assert(this.messages.toArray().length == 0, "Contract state should not be deserialized in @migrate");
+        // retrieve the current state from the contract
+        let raw_vector = JSON.parse(near.storageRead("STATE")).messages;
```
ailisp: Instead of reading from raw state, `migrateState` can takes an argument and `@migrate({})` does deserialization for it. e.g.:

```
    migrateState({messages}: {messages: Vector<OldMessageFormat>}): Vector<NewMessageFormat> {
        let old_messages = messages;
        // same as your implementation below
```

examples/__tests__/test-state-migration.ava.js
```diff
@@ -0,0 +1,36 @@
+import { Worker } from "near-workspaces";
+import test from "ava";
+
+test.beforeEach(async (t) => {
+    const worker = await Worker.init();
+    const root = worker.rootAccount;
+    const contract = await root.devDeploy("./build/state-migration.wasm");
```
ailisp: To make this example align with real world workflow, let's make two contracts: `old-message.wasm` and `state-migration.wasm`. 
In `old-message.wasm`, it has `OldMessageFormat`, `addMessage({ message }: {message: OldMessageFormat})` and `countMessages`. 
And `state-migration.wasm` is just built from current `examples/src/state-migration.ts`.

In the test, developer first deploy `old-message.wasm`, add a couple of old messages. Then he deployed `state-migration.wasm`, call `migrateState` and `countMessages`.

ailisp (commented):
Nice work! The implementation and example is solid. An experienced developer can figure out how to make migration for his contract. I leave a bit improvement to enhance DevX.

examples/src/state-migration.ts
volovyks: Why should we pass old messages?
I think that the whole idea behind migration is to rewrite the data in new format. It means:
- read the old data
- reformat it
- save new data to state

examples/__tests__/test-state-migration.ava.js
volovyks: Good point, refactored to have 2 contracts.

examples/src/state-migration.ts
ailisp: Your approach works and looks intuitive! I was thinking a syntax sugar to read the old data to avoid `JSON.parse(near.storageRead("STATE"))` pattern. User can write:
```
@Migrate
migrateState(oldState: OldStateFormat): NewStateFormat {
    let newState: NewStateFormat = some_reformat_steps(oldState);
    return newState;
}
```
`@Migrate` would know how to load (deserialize) oldState given OldStateFormat. `@Migrate` would also serialize reformat state using newStateFormat.

A possible `@Migrates`'s deserialize:
```
let oldRaw = near.storageRead("STATE")
let old = OldStateFormat.deserialize ? OldStateFormat.deserialize(oldRaw) 
  : OldStateFormat.reconstruct ? OldStateFormat.reconstruct(JSON.parse(oldRaw))
  : JSON.parse(oldRaw)
```

examples/__tests__/test-state-migration.ava.js
ailisp: Great! Test looks self-explanatory!

examples/src/state-migration.ts
ailisp: ^ is i think nice to have in API but it's up to your preference. The PR is considered done!

ailisp (approved):
Thank you! Very helpful pattern and very good implementation

examples/src/state-migration.ts
volovyks: Since we do not have any NEP - or complex approaches are questionable. Let's get back to it if necessary.

